Issue #34, opened 2021-04-27
ImageCube FITS export gives incorrect RA/DEC scale.


iancze:
**Describe the bug**
ImageCube FITS export gives incorrect RA/DEC scale. Somethings's too large by a factor of 1000 or more.

My suspicion is that there's an incorrect [conversion from arcseconds to degrees](https://github.com/MPoL-dev/MPoL/blob/main/src/mpol/images.py#L263).

We should fix this in the code, but also develop a corresponding test for FITS export to make sure we're exporting what we want to be exporting.

trq5014:
When going over the code for ImageCube.to_FITS() I found that there was a conversion from arcseconds to degrees using _arcseconds * 180 / pi_ (conv1). I printed out the **CDELT1** and **CDELT2** values for this conversion and the conversion using _arcseconds / 3600_ (conv2). The results for conv1 were: **CDELT1/CDELT2 = 0.28647889756541** while the results for conv2 were: **CDELT1/CDELT2 = 1.3888... e-06**. The input value for _cellsize_ (the variable used in the arcseconds conversion) was 0.005. Which, if I understand this correctly, is the same value as presented by conv2 (0.005/3600 = CDELT1/CDELT2). I opened both of the resulting fits files in AstroImageJ to check the dimensions. conv1's DEC ranged from +75:57:31.63 to -00:01:148.10 and conv1's RA ranged from 23:59:19.194 to 05:03:48.357 while conv2's DEC ranged from +00:00:04.00 to +00:00:00.00 and conv2's RA ranged from 00:00:00.00 to 00:00:00.266. The image is 800x800 pixels, and each pixel represents 0.005 arcseconds (_cellsize_). Am I misunderstanding something, or shouldn't the fits image's DEC have a range of 4 arcseconds?

iancze:
I'm not sure what your ratios of `CDELT1/CDELT2` are signifying. Are you just saying that they are the same value? They should be, generally CDELT1 == CDELT2

If the image is 800 pixels and each pixel is 0.005 arcseconds, then yes, the image width should be 4 arcseconds wide. It looks like your "conv2" is the correct solution, then? 

If you're satisfied with the solution, it would be great to have this as a pull request back into main (when you're ready w/ tests). Thanks!

trq5014:
Sounds good, and yes I meant it as CDELT1 _and_ CDELT2 = _value_, I see where this may have been confusing.

PR #53, opened 2021-05-18
Fixed Issue #34 - Caused incorrect RA/DEC scale

trq5014:
Changed conversion of arcs to dec in ImageCube.to_FITS() from arcs * 180/pi to arcs/3600
Added test to ensure correct RA/DEC scale in the created fits images

### DIFF
--- a/test/images_test.py
+++ b/test/images_test.py
@@ -1,7 +1,9 @@
 import matplotlib.pyplot as plt
 import pytest
 import torch
+import os
 
+from astropy.io import fits
 from mpol import gridding, images, utils
 from mpol.constants import *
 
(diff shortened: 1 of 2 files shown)


test/images_test.py
```diff
@@ -48,6 +50,26 @@ def test_imagecube_grad(coords):
     # https://github.com/pytorch/pytorch/issues/50014
     loss.backward()
 
+# test for proper fits scale
+def test_imagecube_tofits(coords):
```
iancze, line 54: You might try using the "tmp_path" parameter provided as part of Pytest, and then saving the FITS file to this temporary directory (e.g., https://github.com/MPoL-dev/MPoL/blob/main/test/images_test.py#L108)

The tmp_path is a [Pytest fixture](https://docs.pytest.org/en/6.2.x/fixture.html) that creates a temporary directory for the duration of the run. That way, when the tests are complete, the directory is automatically cleaned up.

test/images_test.py
```diff
@@ -48,6 +50,26 @@ def test_imagecube_grad(coords):
...
+    # sending the basecube through the imagecube
+    imagecube.forward(bcube.forward())
+
+    # creating output fits file with name 'test_cube_fits_file39.fits'
+    # file will be deleted after testing
+    imagecube.to_FITS(fname='test_cube_fits_file39.fits', overwrite=True)
+
+    # inputting the header from the previously created fits file
+    fits_header = fits.open('test_cube_fits_file39.fits')[0].header
+    os.remove('test_cube_fits_file39.fits')
```
iancze, line 70: For example, if you implement the tmp_path, then this remove line wouldn't be necessary.

iancze (changes requested):

test/images_test.py
trq5014: Just updated this and pushed it to my directory. I think everything looks good now!

iancze:
Nice work, thanks!

Issue #13, opened 2021-03-22
Documentation on "packed" image cube format


iancze:
**Is your feature request related to a problem or opportunity? Please describe.**
It would be helpful to have a short description in the documentation visualizing what we mean by "packed" image or visibility cube format. I.e., after np.fft.fftshift has been applied to a normal cube. The fftshift operation works only along the RA and DEC axes to prepare the cube for a 2D FFT. The velocity/frequency axis remains untouched.

For image cubes, we are calling

* "sky_cube" -> the normal orientation of an image on the sky, i.e., for plotting. RA increases to the *left* 
* "packed_cube" -> the application of fftshift(flip(sky_cube)) along the RA and DEC axes. 

For visibility cubes, there is no corresponding "sky," rather just the way we want the array to look on the ground. "u" increases to the right.

This is a rough sketch showing how the cubes are related. 

<img width="827" alt="Screen Shot 2021-05-19 at 8 14 28 PM" src="https://user-images.githubusercontent.com/467948/118900427-efc75280-b8de-11eb-8c1a-55b19bf5e84d.png">

Also probably worthwhile to review the developer documentation on [Units and Conventions](https://mpol-dev.github.io/MPoL/units-and-conventions.html) describing how arrays are packed.

- [ ] why this format is required in the first place (for the FFT conventions)
- [ ] how this is used internally, and what the roles of the `sky_cube` functions are
- [ ] [utils](https://github.com/MPoL-dev/MPoL/blob/main/src/mpol/utils.py) routines for shifting need documentation

For a reference implementation of how we use these packed formats, see the flow of an image cube through [ImageCube](https://github.com/MPoL-dev/MPoL/blob/main/src/mpol/images.py#L152) to a [sky_cube](https://github.com/MPoL-dev/MPoL/blob/main/src/mpol/images.py#L229).

**Describe the solution you'd like**
Write a short `.rst` file either as part of "units and conventions" or near it.

trq5014:
Thank you for updating this! It really helped clear up some of my confusion here. This issue will hopefully be resolved soon now.

PR #55, opened 2021-05-21
Updates to Documentation regarding Issue #13

trq5014:
Added Packed Image Cube section to the units-and-conventions.rst detailing the use of the shift functions and sky_cubes. Also added why this format is required (to the best of my current knowledge)
Updated the documentation of the shifting functions in mpol.utils

I don't expect that all my documentation updates are correct/ presented in the way that was envisioned (Specifically, for the 'Why this format is required' section.  I was a little unsure when learning about this section.). Please let me know which areas need to be fixed. 

I was also thinking it may be useful to include a graphic similar to the one presented in the Issue #13 page. I can create one if you agree on its usefulness.


docs/units-and-conventions.rst
```diff
@@ -118,3 +118,23 @@ If we had a fully sampled grid of :math:`{\cal V}_{u,v}` values, then the operat
...
+
+
+"Packed" Image Cubes
+--------------------
+
+Internally, data is represented by Cubes, or arrays of three dimensions. In the case of MPoL, the dimensions of an MPoL cube represent Declination (DEC, axis one), Right Ascension (RA, axis two), and Velocity/Frequency (axis three). Initially, the Normal Cube, or Sky Cube, is the orientation of an image on the sky. Due to this, the RA increases to the *left*. It is desirable to flip the RA axis to obtain a traditional scaling with RA increasing to the *right*.
+
+We now have a flipped version of the Sky Cube (referred to as the Flip Cube). The next step is to "pack" the cube. In this application, this is carried out by applying ``torch.fft.fftshift()`` to the Flip Cube. This method acts only along the DEC and RA axes (axis one and axis two) leaving axis three untouched and creating a Packed Cube, or Image Cube. This method is represented internally as ``mpol.utils.sky_cube_to_packed_cube()``.
```
iancze, line 128: The Sky Cube itself is kinda tricky. In general, astronomical images are usually plotted with north up and east to the left. "l" = RA and "m" = DEC. So as it appears on the sky, we have the "l" axis increasing to the left. This is a little strange compared to the way we normal look at x,y axis, where x increases to the right.

docs/units-and-conventions.rst
```diff
@@ -118,3 +118,23 @@ If we had a fully sampled grid of :math:`{\cal V}_{u,v}` values, then the operat
...
+
+
+"Packed" Image Cubes
+--------------------
+
+Internally, data is represented by Cubes, or arrays of three dimensions. In the case of MPoL, the dimensions of an MPoL cube represent Declination (DEC, axis one), Right Ascension (RA, axis two), and Velocity/Frequency (axis three). Initially, the Normal Cube, or Sky Cube, is the orientation of an image on the sky. Due to this, the RA increases to the *left*. It is desirable to flip the RA axis to obtain a traditional scaling with RA increasing to the *right*.
+
+We now have a flipped version of the Sky Cube (referred to as the Flip Cube). The next step is to "pack" the cube. In this application, this is carried out by applying ``torch.fft.fftshift()`` to the Flip Cube. This method acts only along the DEC and RA axes (axis one and axis two) leaving axis three untouched and creating a Packed Cube, or Image Cube. This method is represented internally as ``mpol.utils.sky_cube_to_packed_cube()``. 
+
```
iancze, line 129: The FFT operates on a multidimensional array.  For each array direction, as you increase the index into the array, you should be increasing the value of the coordinate. so i=1,2,3... should correspond to larger and larger l. That way the FFT'd array you get out has increasing coordinates for i=1,2,3 for the u coordinate.

docs/units-and-conventions.rst
```diff
@@ -118,3 +118,23 @@ If we had a fully sampled grid of :math:`{\cal V}_{u,v}` values, then the operat
...
+At this point, a two-dimensional Fast Fourier Transformation (described above) is applied to the Image Cube. Now the data has been transformed from a Sky Cube into what we refer to as a Visibility Cube with axes :math:`u,v`. The Visibility Cube is also a Packed Cube similar to the Image Cube, but on the other side of the FFT. By taking this and applying ``torch.fft.fftshift()`` along axes one and two we get a Ground Cube. This method is represented internally as ``mpol.utils.packed_cube_to_ground_cube()``.
```
iancze, line 130: I agree that I think it would be very helpful to have a graphical illustration of this, showing the array at each stage: sky, flipped, packed, with labeled coordinates. I would use the ALMA logo. You should be able to download it here: https://zenodo.org/record/4711811#.YKf3jJNKidY as the logo_cont.fits. 

Ulitmately, you'll want to include these as products built as part of the documentation Makefile and then included/referenced from this rst document. Just like the CSV file with baseline lengths.

docs/units-and-conventions.rst
```diff
@@ -118,3 +118,23 @@ If we had a fully sampled grid of :math:`{\cal V}_{u,v}` values, then the operat
...
+
+At this point, a two-dimensional Fast Fourier Transformation (described above) is applied to the Image Cube. Now the data has been transformed from a Sky Cube into what we refer to as a Visibility Cube with axes :math:`u,v`. The Visibility Cube is also a Packed Cube similar to the Image Cube, but on the other side of the FFT. By taking this and applying ``torch.fft.fftshift()`` along axes one and two we get a Ground Cube. This method is represented internally as ``mpol.utils.packed_cube_to_ground_cube()``. 
+
+To convert from a Ground Cube to a Sky Cube, simply reverse this process. Apply the ``mpol.utils.ground_cube_to_packed_cube()`` function to shift the data to a Packed Visibility Cube, apply the FFT to the Packed Cube to create an Image Cube, then use the function ``mpol.utils.packed_cube_to_sky_cube()`` to obtain a Sky Cube from the Image Cube. 
+
```
iancze, line 133: The visibility cube process is similar, only though it doesn't need a "flip" stage, because we plot u,v like we normally plot x,y. We should also include a figure for this, which would be the corresponding FFT from the ALMA logo.

src/mpol/utils.py
```diff
@@ -6,15 +6,27 @@
```
iancze, line 6: These doc additions are great!

iancze (changes requested):
Thanks for the great start. I added a few suggestions here for next steps. I think it makes sense to try making the example figures using the ALMA logo, and then we can mold the text discussion around those.

docs/units-and-conventions.rst
trq5014: I downloaded the logo_cont.fits file, but the data is 2-dimensional, so I am not sure how to use this as a cube. I could use the .npz file instead if that works?

trq5014:
Just update with a new commit. Let me know if there is anything else that should be changed before this is ready to merge

Issue #35, opened 2021-04-27
Tutorial on how to initialize model with the dirty image


iancze:
**Is your feature request related to a problem or opportunity? Please describe.**
An RML forward model can be optimized from any starting point, but it will make the optimization that much quicker if the starting point already has a low loss value (maximal likelihood). 

One such (easy-to-calculate) image is the dirty image, which already has a maximal likelihood value. However, the dirty image usually contains regions of negative flux, which are not allowed by some model parameterizations (e.g., [BaseCube w/ default Softplus pixel mapping](https://github.com/MPoL-dev/MPoL/blob/main/src/mpol/images.py#L13), which enforces image positivity).

**Describe the solution you'd like**
One way around this is to optimize the model *based off of the image-plane distance* between the dirty image and the state of the ImageCube. The pixel-to-pixel L2 norm seems like a reasonable distance metric in this case.

```
# calculate the dirty image we want to optimize to
# doesn't need to be uniform gridding in this case
# since we aren't using data visibilities just yet
gridder.grid_visibilities(weighting="briggs", robust=0.0)
img = gridder.get_dirty_image("Jy/arcsec^2")
# make img a tensor
dirty_image = torch.tensor(img.copy())

for iteration in range(500):

    optimizer.zero_grad()

    model.forward()
    sky_cube = model.icube.sky_cube

    loss = loss_fn(sky_cube, dirty_image)

    loss.backward()
    optimizer.step()

torch.save(model.state_dict(), "dirty_image_model.pt")
```

The model state can  be saved, and then reloaded right before optimizing the model against the visibility data.

```
model.load_state_dict(torch.load("dirty_image_model.pt"))

# be sure to reset gridded visibilities to "uniform" before optimizing 
gridder.grid_visibilities(weighting="uniform")
dataset = gridder.to_pytorch_dataset()
```

This could be documented as a short, self-contained tutorial, that can be referenced from other tutorials that go all the way to optimizing directly against visibility data.

In short tests so far, this pseudo-dirty image provides a much better optimization starting point than a blank image, which means the learning rate of the optimizer can be tuned more aggressively to reach convergence.

The tutorial should also be clear about the fact that RML optimization doesn't *need to* start from the dirty image, and therefore isn't really a deconvolution algorithm in the same sense as `tclean`.

PR #58, opened 2021-05-28
Tutorial for issue #35, initialize model with dirty image

RCF42:
This is a WIP slightly, I believe I've updated all the documentation and everything else correctly as well. 
Tutorial goes over the steps to get the model optimized based off the dirty image and then saving/loading the model
. Just slightly unsure about certain phrasings and want to make sure my loss function is right. Also need to embed links.

### DIFF
--- a/docs/tutorials/initializedirtyimage.py
+++ b/docs/tutorials/initializedirtyimage.py
@@ -22,11 +22,11 @@
 
 # ## Initializing Model with the Dirty Image
 #
-# In MPol the Regualarized Maximum Likelihood (RML) algorithm is written as an optimization loop (as seen in the [Optimization Tutorial](optimization.html)). With a better starting point for the model parameters, fewer iterations of the loop will need to be done to reach the optimal point. Normally, we just start from a uniform image as seen in the optimization tutorial, but the dirty image is a better starting point because it maximizes the likelihood function.
+# In MPoL, the Regularized Maximum Likelihood (RML) algorithm is written as an optimization loop (as seen in the [Optimization Tutorial](optimization.html)). With a better starting point for the model parameters, fewer iterations of the loop will be needed to reach the optimal point. A default uniform image (as shown in the optimization tutorial) is usually not a great guess as to the final image. A better guess is the dirty image, since it is already a maximum likelihood fit the data (of course, it is unregularized).
 #
-# The problem with the dirty image is that the RML model does not permit negative flux value (see the [BaseCube](../api.html#mpol.images.BaseCube) documentation), while the dirty image contains negative flux values. Overall though, enforcing positivity is ideal for astronomical image reconstruction, because real images do not have negative flux.
+# The problem with the dirty image is that it usually contains negative flux pixels and we'd like to impose the (rather strong) prior that the astrophysical source must have positive intensity values (i.e., no negative flux values are permitted, the implementation is taken care of via the [BaseCube](../api.html#mpol.images.BaseCube) parameterization). So how do we initialize the RML image model to the dirty image if we can't represent negative flux pixels?
 #
-# So this tutorial will demonstrate how to initialize the model paramters to reproduce the dirty image as closely as possible (limited by the enforced positivity of the image) in order to eventually provide a better starting point that can be used by the RML model. We will also go over saving and loading the model. After saving and loading it, it can be then optimized against the visibility data to complete it (though this will not be done in this tutorial). We will use the dataset of the ALMA logo first used in the Gridding and Diagnostic Images tutorial.
+# This tutorial will demonstrate one initialization solution, which is to create a loss function corresponding to the mean squared error between the RML model image pixel fluxes and the dirty image pixel fluxes and then optimize the RML model. We will also cover how to save and load the starting point configuration. After saving and loading it, it can be then optimized against the visibility data to complete it (though this will not be done in this tutorial). We will use the dataset of the ALMA logo first used in the Gridding and Diagnostic Images tutorial.
 #
 #
 
@@ -40,7 +40,7 @@ import matplotlib.pyplot as plt
 from mpol import gridding, coordinates, precomposed, losses, utils
 from astropy.utils.data import download_file
 
-# In this tutorial we will be saving and loading the model we use and so it is important to make sure that when saving and loading a model the cell_size, nchan, and npix are the same between the model and the coordinates if used in another document. More info on the coordinates can be found [here](../api.rst#mpol.coordinates.GridCoords).
+# When saving and loading a model, it is important to make sure that ``cell_size``, ``nchan``, and ``npix`` remain the same. More info on coordinates can be found [here](../api.rst#mpol.coordinates.GridCoords).
 
 # +
 # load the mock dataset of the ALMA logo
@@ -69,12 +69,7 @@ coords = coordinates.GridCoords(
     cell_size=0.03, npix=180
 )  # Smaller cell size and larger npix value can greatly increase run time
 gridder = gridding.Gridder(
-    coords=coords,
-    uu=uu,
-    vv=vv,
-    weight=weight,
-    data_re=data_re,
-    data_im=data_im,
+    coords=coords, uu=uu, vv=vv, weight=weight, data_re=data_re, data_im=data_im
 )
 
 # export to PyTorch dataset
@@ -84,7 +79,11 @@ dset = gridder.to_pytorch_dataset()
 # Show the dirty image
 img, beam = gridder.get_dirty_image(weighting="briggs", robust=0.0, unit="Jy/beam")
 imin, imax = np.amin(img), np.amax(img)
+# -
 
+# Now let's take a look at the dirty image. We've used a different colormap to highlight the many negative flux pixels contained in this image.
+
+# +
 plt.set_cmap(
     "Spectral"
 )  # using Matplotlib diverging colormap to accentuate negative values
@@ -111,14 +110,12 @@ rml.state_dict()  # parameters of the model
 
 # ### Loss Function
 
-# The [loss function](../api.html#module-mpol.losses) that will be used to optimize the initial part of the image is the pixel-to-pixel L2 norm (also known as the Euclidian Norm). It calculates the loss based off of the image-plane distance between the dirty image and the state of the ImageCube in order to make the state of the ImageCube closer to the dirty image.
-#
-# The L2 Norm (Euclidian Norm) is based off of the euclidian distance function and it takes the form of: $ L_{oss} = \sqrt{\sum_{i=1}^{N} (y_i - x_i)^2}  $. [Pytorch provides a loss function for mean squared error](https://pytorch.org/docs/stable/generated/torch.nn.MSELoss.html) which is the squared L2 norm so we will be using the squareroot of that.
+# The [loss function](../api.html#module-mpol.losses) that will be used to optimize the initial part of the image is the pixel-to-pixel L2 norm (also known as the Euclidian Norm). It calculates the loss based off of the image-plane distance between the dirty image and the state of the ImageCube in order to make the state of the ImageCube closer to the dirty image. [Pytorch provides a loss function for mean squared error](https://pytorch.org/docs/stable/generated/torch.nn.MSELoss.html) which is the squared L2 norm so we will be using the squareroot of that.
 
 
 # ### Training Loop
 #
-# Now we start a trainining loop using this loss function to optimize our parameters. There will also be a loss tracker to see how loss falls off over time.
+# Now we train using this loss function to optimize our parameters.
 
 # +
 # %%time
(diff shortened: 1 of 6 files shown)


docs/tutorials/initializedirtyimage.py
```diff
@@ -0,0 +1,190 @@
...
+gridder = gridding.Gridder(
+    coords=coords,
+    uu=uu,
+    vv=vv,
+    weight=weight,
+    data_re=data_re,
+    data_im=data_im,
+)
+
+# export to PyTorch dataset
```
iancze, line 79: I cloned your fork and tried to build the docs on your branch, but got an error here. Your tutorial is written for the new syntax that Brianna created (which is good), but I think your fork still needs to incorporate the upstream changes from MPoL-dev/main. I think you can do this easily using the "Fetch Upstream" changes button on the main page of your forked repo. I'm kind of confused how this builds correctly for you locally when you run `make html`? It seems like it builds correctly in the pull request because those changes are being applied on top of the new syntax changes already in MPoL-dev/main.

docs/tutorials/initializedirtyimage.py
```diff
@@ -0,0 +1,190 @@
...
+
+# + nbsphinx="hidden"
+# %run notebook_setup
+# -
+
+# # Not completely done, still have to do some links do some fleshing out, just want to get it here to make sure there isn't anything critically wrong.
+#
+# ## Initializing Model with the Dirty Image
+#
+# In this tutorial we will be going over how to optimize the model based off of the dirty image. It has a low loss value (maximal likelihood) and so this allows the optimization to be much quicker. The problem with the dirty image is that certain model paramterizations enforce image positivity while the dirty image usually contains regions of negative flux. So in this tutorial we will be optimizing the model to be similar to the dirty image (phrasing could be better here). We will also be going over saving and loading the model. After saving and loading it, it can be then optimized against the visibility data to complete it (though this will not be done in this tutorial). We will be using the dataset of the ALMA logo first used in the Gridding and Diagnostic Images tutorial.
```
iancze, line 27: I would back up a little bit and provide more context. 

* Our RML algorithm is written as an optimization loop (see previous optimization tutorial)
* The number of iterations required to reach the optimal point will be fewer if we have a good starting point for the model parameters
* Currently we're just starting from a flat (uniform) image. The dirty image is a better starting point, since we've already established that it maximizes the likelihood function (if not priors)
* We've implemented an RML model that does not permit negative flux values. This is a good thing for astronomical image reconstruction, because real images don't have negative flux. 
* However, this presents a problem for using the dirty image as a starting point, since it contains negative values.
* This tutorial shows how to initialize the model parameters to reproduce the dirty image as closely as possible, under the constraint that all pixels must be positive

docs/tutorials/initializedirtyimage.py
```diff
@@ -0,0 +1,190 @@
...
+    data_re=data_re,
+    data_im=data_im,
+)
+
+# export to PyTorch dataset
+dset = gridder.to_pytorch_dataset()
+# -
+
+# Show the dirty image
+img, beam = gridder.get_dirty_image(weighting="briggs", robust=0.0, unit="Jy/beam")
```
iancze, line 84: It would be useful to accentuate the negative values in the dirty image using a diverging colormap: https://matplotlib.org/stable/tutorials/colors/colormaps.html

docs/tutorials/initializedirtyimage.py
```diff
@@ -0,0 +1,190 @@
...
+# The results are seen here with the state_dict() having changed and the image now more closely resembling the dirty image, though the dirty image contains negative values which this does not so they can never be completely the same.
+
+fig, ax = plt.subplots(nrows=1)
+im = ax.imshow(
+    np.squeeze(rml.icube.sky_cube.detach().numpy()),
+    origin="lower",
+    interpolation="none",
+    extent=rml.icube.coords.img_ext,
+)
+plt.colorbar(im)
```
iancze, line 164: We can use the same colorbar as the dirty image to show that there are no negative values in this representation

docs/tutorials/initializedirtyimage.py
```diff
@@ -0,0 +1,190 @@
...
+
+
+# ### Loss Function
+
+# The loss function that will be used to optimize the initial part of the image is the pixel-to-pixel L2 norm (also known as the Euclidian Norm). It calculates the loss based off of the image-plane distance between the dirty image and the state of the ImageCube in order to make the state of the ImageCube closer to the dirty image.
+#
+# The L2 Norm (Euclidian Norm) is based off of the euclidian distance function and it takes the form of: $ L_{oss} = \sqrt{\sum_{i=1}^{N} (y_i - x_i)^2}  $
+
+
+def l2norm(sky, dirt):
```
iancze, line 112: This loss function works, but I think it would be simpler to just use `loss_fn = torch.nn.MSELoss()`

docs/tutorials/initializedirtyimage.py
```diff
@@ -0,0 +1,190 @@
...
+
+    loss_tracker.append(loss.item())
+    loss.backward()
+    optimizer.step()
+# torch.save(rml, "dirty_image_model.pt")
+torch.save(rml.state_dict(), "dirty_image_model.pt")
+# -
+
+# We also can save the state of the model or the entire model itself. The torch.save(model.state_dict(), PATH) will just save the state.dict() which are the model's learned parameters. The alternative will save the entire model.
+# https://pytorch.org/tutorials/beginner/saving_loading_models.html
```
iancze, line 146: I think we want to steer the user towards saving `model.state_dict` over the entire model, since state dict is a much simpler object to understand.

docs/tutorials/initializedirtyimage.py
```diff
@@ -0,0 +1,190 @@
...
+plt.colorbar(im)
+
+rml.state_dict()
+
+# ### Loading the Model
+#
+# To demonstrate the saving and loading of the model here we will be resetting the parameters of the model and then reloading them.
+
+rml = precomposed.SimpleNet(coords=coords)
+rml.state_dict()  # parameters of the model
```
iancze, line 173: These are the *uninitialized* parameters of the model, or with the default initializations

docs/tutorials/initializedirtyimage.py
```diff
@@ -0,0 +1,190 @@
...
+
+# To continue and optimize our model against the visibility data the gridded visibilitied need to have uniform weighting
+
+gridder._grid_visibilities(weighting="uniform")
+img, beam = gridder.get_dirty_image(weighting="uniform")
+dataset = gridder.to_pytorch_dataset()  # creating a new dataset
+
+# ### Conclusion
+#
+# This tutorial shows how to work towards a better starting point for RML optimization by using an image with maximal likelihood but RML optimization does not need to start from here and can start from a blank image. This is unlike a deconvolution algorithm such as the CASA tclean. The training loop done here is just the set up and the model/image still needs further regularization and optimization against the visibility data.
```
iancze, line 188: You can link to the optimization and cross validation tutorials for more information

iancze (changes requested):
This looks like a very nice start. I have some suggestions for colorbar plots. 

Also, it would be helpful if you could pull upstream changes to your local fork. I can't seem to build the documentation locally, so I'm just editing things based off of what I see in the source code here.

docs/tutorials/initializedirtyimage.py
```diff
@@ -0,0 +1,190 @@
...
+fig, ax = plt.subplots(ncols=1)
+snp = ax.imshow(np.squeeze(img), **kw)
+ax.set_title("image")
+ax.set_xlabel(r"$\Delta \alpha \cos \delta$ [${}^{\prime\prime}$]")
+ax.set_ylabel(r"$\Delta \delta$ [${}^{\prime\prime}$]")
+plt.colorbar(snp)
+
+# ### Model and Optimization Setup
+#
+# Here we will be setting up the optimizer and the image model (RML). If this is unfamiliar please reference the Optimization tutorial. The initial parameters of the model are also displayed and can be contrasted with them after the optimization loop. }
```
briannazawadzki, line 95: I think it would be nice to include a link to the optimization tutorial here. Also, this may be a bit nitpicky, but I think you should consider rephrasing the tutorial to use more direct language, e.g. "Here we will be setting up the optimizer" --> "Here we set up the optimizer"

docs/tutorials/initializedirtyimage.py
```diff
@@ -0,0 +1,190 @@
...
+rml.state_dict()
+
+# ### Loading the Model
+#
+# To demonstrate the saving and loading of the model here we will be resetting the parameters of the model and then reloading them.
+
+rml = precomposed.SimpleNet(coords=coords)
+rml.state_dict()  # parameters of the model
+
+# Here you can clearly see the state_dict() returning to its original state from the beginning of the tutorial and then once we reload it, it reverts to the values it was when we saved it.
```
briannazawadzki, line 175: Can you rephrase this for clarity? I know what you mean, but I think it may be a bit confusing for a reader who is new to this.

docs/tutorials/initializedirtyimage.py
```diff
@@ -0,0 +1,190 @@
...
+
+rml.load_state_dict(torch.load("dirty_image_model.pt"))
+rml.state_dict()  # parameters of the model
+
+# To continue and optimize our model against the visibility data the gridded visibilitied need to have uniform weighting
+
+gridder._grid_visibilities(weighting="uniform")
+img, beam = gridder.get_dirty_image(weighting="uniform")
+dataset = gridder.to_pytorch_dataset()  # creating a new dataset
```
briannazawadzki, line 184: I'm not sure what purpose these three lines serve. You already called `get_dirty_image` and `to_pytorch_dataset` earlier in the tutorial. I don't think you should call `_grid_visibilities` in the tutorial at all now since the user is no longer meant to access that function directly. Using `get_dirty_image` and/or `to_pytorch_dataset` should already grid the visibilities in the way the user needs.

docs/tutorials/initializedirtyimage.py
RCF42: I made sure I had done fetch origin and everything before doing the pull request (or at least I thought I did) and when I do make html the documentation was all building properly for me. I'll try and see what's up with that.

iancze: I think fetch origin will just update your local repository with the changes to RCF42/MPoL, which should be none, since you were the only one making changes. What you want to do is update RCF42/MPoL with the new changes that are in MPoL-dev/MPoL. I think you can do this using the "Fetch Upstream" button I mentioned. See also here: https://docs.github.com/en/github/collaborating-with-pull-requests/working-with-forks/merging-an-upstream-repository-into-your-fork

RCF42: Ah okay, my bad

RCF42: So I've done the fetch upstream and made sure that my branch is up to date with the overall main but I'm facing a problem that when I try and build the documentation it will make just about everything fine but when it comes to the tutorials it gives me:
 `E:\robert\Documents\GitHub\MPoL\docs\index.rst:33: WARNING: toctree contains reference to nonexisting document 'tutorials/gridder'`
for every single tutorial and nothing relating tutorials seems to appear under the build folder. Even if I run it off of my main branch (which is even with this one's) the problem still occurs.

iancze: Is it possible this is a local pip install issue, as per this [discussion](https://github.com/MPoL-dev/MPoL/discussions/66#discussioncomment-814423)?

RCF42: I don't think so because with pip list MPoL appears the way you said it should and both Sphinx and jupytext are also there. I have checked my local branch and I also do have that commit and gridded visibilities is _grid_visibilities. I do get some other errors when doing make html (like it says [Errno 2] No such file or directory: '_static/baselines/build/baselines.csv'.) but none of them appear the error about the 'nonexistent' tutorials.

iancze: What happens if you do

```
$ python
>>> import mpol 
>>> mpol.__version__
>>> from mpol.gridding import Gridder
>>> dir(Gridder)
```
?

This should show the updated routines, like 

```['__class__', '__delattr__', '__dict__', '__dir__', '__doc__', '__eq__', '__format__', '__ge__', '__getattribute__', '__gt__', '__hash__', '__init__', '__init_subclass__', '__le__', '__lt__', '__module__', '__ne__', '__new__', '__reduce__', '__reduce_ex__', '__repr__', '__setattr__', '__sizeof__', '__str__', '__subclasshook__', '__weakref__', '_fliplr_cube', '_grid_visibilities', '_grid_weights', '_histogram_channel', '_histogram_cube', '_null_dirty_beam', 'get_dirty_beam', 'get_dirty_beam_area', 'get_dirty_image', 'sky_vis_gridded', 'to_pytorch_dataset']```

Also, are there any build errors that happen before the one you listed?

You might also try doing a ``make clean`` before you restart ``make html``, just in case stale files are messing something up.

RCF42: All the updated routines appear and there are no build errors that happen before the one I listed. I have been doing the make clean to make html as well. I've also done fresh installs of it all and still get the same problem. I'll see what happens when I switch what computer I'm working on, hopefully that will work.

docs/tutorials/initializedirtyimage.py
RCF42: I redo the uninitialized parameters to show the difference between them and then reloading them, though displaying that isn't too important. Though I suppose I should change the comments to make that clearer.

iancze: Yup, that makes sense.

docs/tutorials/initializedirtyimage.py
RCF42: I've tried it completely fresh from a new computer and it still didn't work so I'm mildly at a lost for why the problem's happening.

hgrzy: In your directory and virtual machine if you do the equivalent of "pip show jinja2" for your system, do you get version 3.0.1 or version 2.11?

iancze: After talking with Robert, I think his main issue stems from how `make` works (or doesn't) on windows. But I did notice this error when I was installing a fork recently. The strange thing is that it appears to be transient... I think it's nbsphinx that relies upon jinja2 2.11, but sometimes the latest jinja2 ends up being installed by accident. I'm not really sure how this is possible with pip, that the dependencies are resolved non-deterministically. @hgrzy have you come across this error too, and if so, was it semi-random?

hgrzy: Yes, I was getting that error a week or two ago, but I just tried replicating it and it seems to install the jinja2 version that is desired even though in both cases I was using pip version 20.2.3. I saved the terminal as a text file when I was getting the error though and it always occurred in the the process of installing the developer package at the point where it's running the setup.py files, though it isn't any particular setup.py file. It also included this in the output:

`ERROR: After October 2020 you may experience errors when installing or updating packages. This is because pip will change the way that it resolves dependency conflicts.

We recommend you use --use-feature=2020-resolver to test your packages with the new resolver before it becomes the default.`

hgrzy: I just looked at the release history for nbsphinx. As of 10 hours ago pip install nbsphinx now installs version 0.8.6 (which seems to run jinja2 version 3.0.1 fine) instead of 0.8.5 (requires jinja2 version 2.11.0) which is probably why I cannot replicate the error. So maybe this is no longer an issue?

iancze: Good detective work! Hopefully this is resolved for future builds---it seems specifically an nbsphinx problem. If it crops up again we'll need to pin the jinja2 version, which I'd like to avoid doing unnecessarily.

Issue #26, opened 2021-04-14
Getting started with PyTorch tutorial


iancze:
**Is your feature request related to a problem or opportunity? Please describe.**
A small, self-contained tutorial that introduces PyTorch to astronomers that may be unfamiliar with it.

**Describe the solution you'd like**
This would mostly follow the arc of other introductory tutorials (e.g., [PyTorch docs intro](https://pytorch.org/tutorials/beginner/basics/intro.html), [Stanford CS230](https://cs230.stanford.edu/blog/pytorch/)) introducing and discussing 
* Initialization of tensors
* Calculation of gradients
* The idea of gradient descent (without yet introducing the PyTorch optimizer object)

These could be built up around the idea of minimizing an, easily plotted, 1D function whose gradient can be caclulated analytically. This could be a polynomial with a clear global minimum and no local minima in the search space.

The main purpose is to assuage the fears of new users who are familiar with Python, might be interested in using MPoL, but would see the unfamiliar requirement of PyTorch as a potential barrier. The goal is to show users that using MPoL is just like writing Python. At the end of the tutorial, links to numerous other PyTorch introductory materials will be provided.

**Describe alternatives you've considered**
We've considered just linking out introductory material to existing PyTorch guides, but I think there is enough mismatch in scope and tone that many new MPoL users would find this disorienting (e.g., how does this apply to RML imaging?!?) and lose interest in MPoL. After landing on the documentation page, we want them to quickly understand that this RML package is accessible to normal (astronomical) Python users.

iancze:
Closed by #60

PR #60, opened 2021-05-31
PyTorch Introduction Tutorial for issue #26

hgrzy:
This is a tutorial including: initialization of tensors, calculation of gradients, and the idea of gradient descent. 

I used Base64 encoding to include some images. There may be a better way to do this.


docs/tutorials/PyTorch.py
```diff
@@ -0,0 +1,322 @@
...
+# Let's define some variable y in terms of x:
+
+y = x ** 2
+
+#  We see that the value of ``y`` is as we expect---nothing too strange here.
+
+print(f"y: {y}")
+print(f"x: {x}")
+
+# But what if we wanted to calculate the gradient of ``y`` with respect to ``x``? Using calculus, we find that the answer is dy/dx = 2*x. The derivative evaluated at ``x=3`` is ``6``.
```
iancze, line 80: We should pick a uniform style for math equations here. I just put ``x`` and ``y``, but maybe it makes $x$.

docs/tutorials/PyTorch.py
```diff
@@ -0,0 +1,322 @@
...
+#
+# We can see if PyTorch gets the same answer as us if we do:
+
+y.backward()  # populates gradient (.grad) attributes of y with respect to all of its independent variables
+x.grad  # returns the grad attribute (the gradient) of y with respect to x
+
+# PyTorch uses the concept of automatic differentiation to calculate the derivative. Instead of computing the derivative as we would by hand, the program is using a computational graph.
```
iancze, line 87: I removed the detailed notes on graphs and nodes. But it might be worth having a few sentences on that this is basically the mechanistic application of the chain rule.

docs/tutorials/PyTorch.py
```diff
@@ -0,0 +1,322 @@
...
+# We will choose `step size = 0.1`:
+
+# +
+t = np.linspace(-5, 5, 100)
+plt.plot(t, t ** 2)
+
+
+plt.scatter(-4, 16)  # point
+
+plt.scatter(-3.2, 10.24)  # second point
```
iancze, line 150: How are the values computed in these points? Would it be possible to compute them directly, rather than hard-coding them?

docs/tutorials/PyTorch.py
```diff
@@ -0,0 +1,322 @@
...
+plt.show()
+# -
+
+# We would continue this process of checking if the gradient is nearly zero, and taking a step in the direction of steepest descent until we reach the bottom of the valley. We'll say we've reached the bottom of the valley when the absolute value of the gradient is <0.1:
+
+# +
+t = np.linspace(-5, 5, 100)
+plt.plot(t, t ** 2)
+
+x_vals = np.array(
```
iancze, line 188: How are the values computed in these arrays? Would it be possible to compute them directly, rather than hard-coding them?

iancze (changes requested):
I made a few changes trimming down the scope, like we discussed. I also deleted the files from the tensorflow (docs/tutorials/runs) that were accidentally committed to the repo.

docs/tutorials/PyTorch.py
briannazawadzki: Echoing this suggestion because I think the hard-coded values 1) make the process more opaque and 2) quickly become a bit confusing/overwhelming to look at, especially in a tutorial where someone is presumably trying to understand the basics.

docs/tutorials/PyTorch.py
```diff
@@ -0,0 +1,322 @@
...
+)
+
+plt.scatter(x_vals, y_vals)
+
+plt.xlim(xmin=-5, xmax=5)
+plt.ylim(ymin=0, ymax=25)
+plt.show()
+# -
+
+# This works, but it takes a long time since we have several small steps. We could speed up the process by taking large steps. Starting at (-4,16) and increasing the step of size to 1.5 we find:
```
briannazawadzki, line 246: I think you should de-emphasize (-4, 16) and just say something like "starting at the same initial location" since you are trying to highlight how changing the step size affects the process.

docs/tutorials/PyTorch.py
```diff
@@ -0,0 +1,322 @@
...
+
+x_vals = np.array([-4, 8, -16])
+y_vals = np.array([16, 64, 256])
+
+plt.plot(x_vals, y_vals)
+
+
+plt.xlim(xmin=-20, xmax=20)
+plt.ylim(ymin=0, ymax=260)
+plt.show()
```
briannazawadzki, line 309: Since you went to the trouble of saving qp, yp, qp2, yp2, etc., I think you should use them in `x_vals` and `y_vals` instead of hard-coding the values. Otherwise it's more difficult to see exactly where these numbers are coming from.

I also think this plot needs more description, since I don't see an explanation of what `plt.plot(x_vals, y_vals)` is plotting. I think it's pretty clear what you're plotting here, but adding a sentence explicitly stating that is probably good practice.

docs/tutorials/PyTorch.py
```diff
@@ -0,0 +1,322 @@
...
+# We wouldn't be able to see all the way to the bottom of the valley, but we could feel which way is down based on where we are standing. We would take steps in the downward direction and we'd know when to stop when the ground felt flat.
+#
+# One other thing we'd have to consider is our step size. If we take very small steps in the direction of the descent, it will take us a longer time than if we take larger steps. However, if we take super long steps, we might completely miss the flat part of the valley, and start ascending the other side of the valley.
+
+# We can look at the gradient descent from a more mathematical lense by looking at the graph z = t ** 2:
+
+t = np.linspace(-5, 5, 100)
+plt.plot(t, t ** 2)
+
+# We start on the left hill at the point (-4, 16):
```
briannazawadzki, line 102: Have you considered starting at a random point? I think that might make it easier for you to write this in a general way rather than hard-coding so many values.

hgrzy:
For the automation, I reworked the first step to have more clear formatting with the equation given in the markup. Then after that is plotted I made an automated process with a loop. I thought it may be useful to see one step done without the whole while loop going on, but if it seems redundant I can remove it. 

I also considered using panda to display a chart with x and y coordinates with the slope at each point, but I thought it might be too overwhelming so I left it out.

Let me know if you have any thoughts!

docs/tutorials/PyTorch.py
```diff
@@ -0,0 +1,346 @@
...
+plt.plot(x, y(x))  # plot y = x ** 2
+
+# We chose step size of 0.1
+step_size = 0.1
+
+# Define a function to get slope of y = x ** 2 at any x value:
+def dy_dx(x_input):
+    gradient = 2 * x_input  # derivative of y = x ** 2
+    return gradient
+
```
iancze, line 167: Would it be possible to use PyTorch directly to calculate the gradients here? Either instead of the analytical gradient, or alongside it? I think that might make the code a bit shorter and underscore the point of why we're using autodifferentiation for optimization.

docs/tutorials/PyTorch.py
iancze: I think it's OK to start at a pre-chosen point, since it looks visually pretty nice. If you were to choose a random point, I think you'd want to first set the random seed to a constant, such that the tutorial is repeatable each time. And then if you're doing this tuning, you might as well just choose a point in the first place.

hgrzy:
Thank you for the suggestion! I omitted the dy_dx function and replaced those operations with PyTorch grad attributes. I had to add some explanation to the use of .item() and lengthened some of the processes due to the nature of getting gradients with PyTorch, but I think it addresses the purposes of why we're using autodifferentiation.

Issue #37, opened 2021-04-27
Estimate visibility variance within each cell


iancze:
**Is your feature request related to a problem or opportunity? Please describe.**
One challenge with working with real data is that the uncertainties (weights) on the visibility data might not be correctly calibrated, meaning that they are not an accurate description of the data-generating process, and thus the statistical interpretation of the loss function is compromised.

Within the CASA framework, one way to look at this is by calculating the scatter of the residuals in units of sigma, where `sigma = sqrt(1/weight)`. This requires a good model image, which can frequently be provided by tclean with the `savemodel='modelcolumn` option. 

`scatter = (data - model)/sigma`. For dual polarization datasets, this calculation can be done for the real and imaginary components of the XX and YY visibilities. A well behaved dataset (bandpass calibrator for 2018.1.00312.S) will have the following properties

<img width="568" alt="correct" src="https://user-images.githubusercontent.com/467948/118566128-6dedf280-b741-11eb-96bc-969a0a94c5a3.png">

where the residuals (mostly) approximate the expected Gaussian. By contrast, a poorly behaved dataset (DSHARP HD 143006 continuum) looks like 

<img width="578" alt="mismatched" src="https://user-images.githubusercontent.com/467948/118566186-84944980-b741-11eb-8a36-7daa54ea6512.png">

In fact, this dataset's sigma values are *underpredicted* by a factor of sqrt(2), which means the weights are too large by a factor of 2. For measurement sets that are comprised of multiple execution blocks across several ALMA cycles, it's possible for one cycle's weights to be correct and another cycle's weights to be scaled incorrectly (DSHARP AS 209 continuum). So this really is something that the user should check carefully on an execution block-by-execution block basis.

It's also possible that some datasets will contain outlier visibilities, like in the following plot 

<img width="585" alt="outliers" src="https://user-images.githubusercontent.com/467948/118566343-b73e4200-b741-11eb-82db-d3f1daae53c9.png">

Normally these bad visibilities should be flagged by the CASA reduction process, but sometimes a user might forget to export these flags when they are reading the raw visibilities from the measurement set (like I have...). 

**Describe the solution you'd like**
The best advice is for the user to thoroughly inspect the weights of their dataset using CASA tools before exporting their visibilities to MPoL, since this is a reduction bug/problem inherent to CASA and is easily (enough) identified using the tclean steps discussed above. We will need to add some sort of documentation about this, perhaps to the [visread](https://github.com/MPoL-dev/visread) package. 

[...]

PR #71, opened 2021-06-04
[WIP] to calculate visibility standard deviation within each cell

iancze:
When complete, pull request will address #37.

Additional things needed:

* incorporate this estimate as an optional check (via parameter arg, like `check_visibility_scatter=True/False` or something) that is part of the to_pytorch_dataset() and get_dirty_image routines. Such that if any cell has a scatter greater than `scatter_threshold`, raise an error. Allow `scatter_threshold` to be a configurable parameter to the routine. Also tests exploring these logic branches.
* update the gridding tutorial to showcase this functionality.


src/mpol/gridding.py
```diff
@@ -559,7 +559,7 @@ def get_dirty_image(
 
         return img.real, self.get_dirty_beam()
 
-    def to_pytorch_dataset(self):
+    def to_pytorch_dataset(self, check_visibility_scatter=True, max_scatter=np.sqrt(2)):
```
iancze, line 562: sqrt(2) might be a little bit too large as a default argument (since some CASA routines are off by precisely this factor). What about 20%, e.g. 1.2?

briannazawadzki:
Okay, I've addressed the remaining things needed, but I wouldn't mind another set of eyes on it before we merge to main!

src/mpol/gridding.py
briannazawadzki: Oops, just saw this. Sure, I'll change that.

iancze:
Looking good! The addition of this functionality actually highlights yet another problem with the CASA weights. I am pretty sure that the weights of the mock ALMA logo dataset are incorrect! Meaning that the synthetic visibilities produced from simobserve itself have the incorrect weights/sigma! Therefore many of the simpler routines using the `mock_dataset` fixture are erroring. I'll update the mock dataset and then we can focus in on simplifying the tests.

I think we might want to move `_check_scatter` inside of the `_grid_visibilities` routine too.

Issue #61, opened 2021-06-01
HD 143006 Imaging Tutorial Part 1


iancze:
An overview of the HD 143006 imaging series is provided in #25 

Part I should cover 

* A paragraph describing the idea of RML
* A paragraph describing how to get the data, and specific processing concerns. The actual data preparation will be done as part of [mpoldatasets](https://github.com/MPoL-dev/mpoldatasets) and [visread](https://github.com/MPoL-dev/visread), but @iancze will eventually summarize with a brief description here:
    - Download continuum data from DSHARP site (visread) 
    - Plotting and examine DSHARP CLEAN FITS (visread)
    - Extract visibilities using CASA table tools (visread)
    - Examine whether weights are scaled correctly (visread)

So the main contents of this tutorial will be
* Basic image setup, imports, data loading for MPoL.  Processed `.npz` file available for download [here](https://zenodo.org/record/4904794/files/HD143006_continuum.npz), which can follow the `astropy` loading in other tutorials.
* A plot of the u,v positions of the visibilities
* Make dirty image using gridder and compare this to the FITS image produced by the DSHARP survey. FITS image [available here](https://almascience.eso.org/almadata/lp/DSHARP/) for download.

hgrzy:
Is the main focus to compare DSHARP with the dirty image (i.e. that's what we should focus on explaining)? I know a lot of the code is very similar to some of the smaller tutorials and I want to make sure the scope of this tutorial is wide enough to get the point across but without running into any problems of duplicating too much of the smaller tutorials.

iancze:
I think the main goal of this part I tutorial is to familiarize the reader with the DSHARP dataset. The code for gridding, producing the dirty image, and displaying it will indeed be familiar with the other tutorials. 

The comparison with the DSHARP FITS image is basically to show that our MPoL-produced dirty image looks similar to the actual FITS CLEAN image provided by the survey. (Of course, the MPoL dirty image still has sidelobe responses because it hasn't been "cleaned").

hgrzy:
Okay thank you!

iancze:
Closed by #74

PR #74, opened 2021-06-11
Created HD143006 tutorial part 1

trq5014:
Added file HD143006_Tut_P1 to docs/tutorials folder
Modified Makefile and index.rst for the addition of the new tutorial
Collaboration on tutorial with: @trq5014, @hgrzy, @RCF42
Tutorial to close Issue #61


trq5014:
Notebook is not currently building correctly as an HTML, attempting to fix this now

docs/index.rst
```diff
@@ -40,6 +40,7 @@ If you'd like to help build the MPoL package, please check out the :ref:`develop
    tutorials/crossvalidation
    tutorials/gpu_setup.rst
    tutorials/initializedirtyimage
+   tutoriala/HD143006_Tut_P1
```
hgrzy, line 43: "tutoriala" --> "tutorials"

hgrzy:
I believe it's not passing the doc test because it's timing out when downloading the npz file. It took 15 or so minutes for me to locally run make html but it does create the index.html file with this version of the tutorial

iancze:
Thanks! I'm going to take a look at it now. It looks like the timeout might be some sort of web error, perhaps originating from the server the FITS file is hosted on. I'll take a closer look.

trq5014:
So after some minor changes (changing the download from the original alma site to a dropbox link temporarily), the docs test was reran and failed again due to the kernel shutting down: 
image
 The dataset we are using is rather large and building our tutorial takes approx 5-8 minutes and a lot of computing power. Should this be changed from a .py file to a .rst file instead? Other possible solutions would be to do something similar to the randomly drawn 9/10 dataset from the [cross validation tutorial](https://mpol-dev.github.io/MPoL/tutorials/crossvalidation.html) to limit the size of the data we are actually using here. Let me know what you think

trq5014:
The current runtime (on my machine) for the script is: 
I was able to get this down to 305 seconds by using 1/10th of the dataset when plotting the U,V Visibilities
I was able to get this down further to 226 seconds by using 4/5ths of the data set when creating the gridder object which merited very similar resulting images.

I think that at a minimum it is worth using 1/10th of the dataset or less to plot the U,V visibilities since it results in the exact same plot, and we can reference this section of the cross validation tutorial. Let me know what you think and I will implement one of these changes.

docs/tutorials/HD143006_Tut_P1.py
```diff
@@ -0,0 +1,164 @@
...
+    plt.xlim(left=-.75, right=.75)
+    plt.ylim(bottom=-.75, top=.75)
+    return ax
+
+plot(img, savename="uniform.png", imtitle="uniform")
+plot(img1, savename="robust_1.0.png", imtitle="robust_1.0")
+plot(img2, savename="robust_0.png", imtitle="robust_0")
+plot(img3, savename="robust_-1.0.png", imtitle="robust_-1.0")
+
+# Below we plot the DSHARP CLEAN image alongside the MPoL RML Dirty Image weighted with the Briggs scale and `robust=0.0` for comparison.
```
iancze, line 144: Just "dirty image," not RML dirty image.

docs/tutorials/HD143006_Tut_P1.py
```diff
@@ -0,0 +1,164 @@
...
+ax[0].set_ylim(bottom=-.75, top=.75)
+ax[0].set_xlabel(r"$\Delta \alpha \cos \delta$ [${}^{\prime\prime}$]")
+ax[0].set_ylabel(r"$\Delta \delta$ [${}^{\prime\prime}$]")
+ax[1].imshow(np.squeeze(img2), **kw)
+ax[1].set_title('MPoL RML Dirty Image')
+ax[1].set_xlim(left=-.75, right=.75)
+ax[1].set_ylim(bottom=-.75, top=.75)
+ax[1].set_xlabel(r"$\Delta \alpha \cos \delta$ [${}^{\prime\prime}$]")
+ax[1].set_ylabel(r"$\Delta \delta$ [${}^{\prime\prime}$]")
+plt.tight_layout()
```
iancze, line 160: It would be helpful to play with the plotting to see if we can get these figures to show up larger, perhaps as a two-row figure? The specific format of the documentation site has a pretty narrow horizontal space.

docs/tutorials/HD143006_Tut_P1.py
```diff
@@ -0,0 +1,164 @@
...
+
+
+# As you can see, the Dirty Image created by MPoL looks very similar to the image the DSHARP team came up with using the CLEAN algorithm. While the RML Dirty Image appears more noisy, it also appears sharper and features are more distinct. Using a very simple process with very little data processing MPoL is able to produce an image of comparable quality to the final product of the CLEAN algorithm. In the next part of the HD143006 tutorial, we will be cleaning the image and explaining how we can get an even better result using Neural Networks, Optimization, and Cross Validation with help from [PyTorch](pytorch.org).
```
iancze, line 163: I would revise the text to be a bit more formal w.r.t. to the DSHARP image, and cite [Andrews et al. 2018](https://ui.adsabs.harvard.edu/abs/2018ApJ...869L..41A/abstract).

We haven't done any RML imaging yet at this point, we've just compared the dirty image produced by MPoL (mostly as a diagnostic functionality) to the "official" tclean image produced by the DSHARP survey using CASA / tclean.

iancze (changes requested):
This is a very nice start! Thank you for the excellent work @trq5014 @hgrzy @RCF42 . I mainly have a few comments about tightening up/polishing the language in places, making the distinction clear between a dirty image and RML imaging, and plotting conventions for RA. 

I wouldn't worry too much about the errors on Github Actions at this point. I'll keep digging into what's going on, I'm confident we can find a solution somehow, probably with caching files.

docs/tutorials/HD143006_Part_1.py
```diff
@@ -0,0 +1,201 @@
...
+
+cdelt_scaling = dfits[0].header['CDELT1'] * 3600 # scaling [arcsec]
+cell_size = abs(cdelt_scaling) # [arcsec]
+# closing fits file
+dfits.close()
+
+from mpol import gridding, coordinates
+
+# creating Gridder object
+coords = coordinates.GridCoords(cell_size=cell_size, npix=3000)
```
trq5014, line 140: Should i change this to 512? I thought it should be the same size as the fits image dimensions but this creates a large memory need.

iancze: 3000 is indeed too large---the DSHARP FITS files are very large just to fill the full primary beam. We don't need to do that for the purposes of the tutorial. So long as you have enough pixels to comfortably fit the image. 512 might be enough, or maybe you'll need to go to 800 or 1024.

trq5014: Sounds good, I will try a few out

iancze:
I think the Github Actions error on docs build comes from exceeding the RAM limit, which is 7Gb on linux machines. When I ran the tutorial build locally this easily exceeded that (12Gb resident memory). I think this is probably coming from the invocation of the gridding commands. 

We could go through the tutorial with a memory profiler to isolate the offending steps (e.g., https://pypi.org/project/memory-profiler/). I have a feeling it's the 

```
img, beam = gridder.get_dirty_image(weighting='uniform')
img1, beam1 = gridder.get_dirty_image(weighting="briggs", robust=1.0, unit="Jy/arcsec^2")
img2, beam2 = gridder.get_dirty_image(weighting="briggs", robust=0.0, unit="Jy/arcsec^2")
img3, beam3 = gridder.get_dirty_image(weighting="briggs", robust=-1.0, unit="Jy/arcsec^2")
```
lines. If so, there's probably a way we could rewrite the `get_dirty_image` routine to be more memory efficient.

trq5014:
I changed these lines around a bit on my local branch to see if it would help and I also ran the memory_profiler and have attached the results as a text file here. 
[mem_profiler_results.txt](https://github.com/MPoL-dev/MPoL/files/6671631/mem_profiler_results.txt)

docs/tutorials/HD143006_Part_1.py
```diff
@@ -0,0 +1,211 @@
...
+)
+
+# downloading extracted visibilities file
+fname_EV = download_file(
+    'https://zenodo.org/record/4904794/files/HD143006_continuum.npz',
+    cache=True,
+    pkgname='mpol',
+)
+
+# Now that we have the files, let us examine the fits image created by the DSHARP team using the CLEAN algorithm ([Andrews et al. 2018](https://ui.adsabs.harvard.edu/abs/2018ApJ...869L..41A/abstract)). First let's extract our data and then, to properly display the image, we will need to change from units of pixels to arcseconds. To do this, we will calculate the image extent.
```
iancze, line 60: The Andrews et al. citation is for the DSHARP survey in particular, but this reads as though Andrews et al. is the reference for the CLEAN algorithm.

When used to describe the [FITS files](https://en.wikipedia.org/wiki/FITS), the acronym is usually capitalized.

docs/tutorials/HD143006_Part_1.py
```diff
@@ -0,0 +1,211 @@
...
+# 
+
+# load extracted visibilities from npz file
+dnpz = np.load(fname_EV) 
+uu = dnpz['uu']
+vv = dnpz['vv']
+weight = dnpz['weight']
+data = dnpz['data']
+
+# Let's quickly plot the U,V visibilities as seen in the [Cross Validation Tutorial](https://mpol-dev.github.io/MPoL/tutorials/crossvalidation.html) and the [Visread docs](https://mpol-dev.github.io/visread/tutorials/introduction_to_casatools.html#Get-the-baselines).
```
iancze, line 110: plot the $u,v$ coordinates of the visibilities

docs/tutorials/HD143006_Part_1.py
```diff
@@ -0,0 +1,211 @@
...
+
+fig, ax = plt.subplots(nrows=1)
+ax.scatter(uu,vv, s=.5, rasterized=True, linewidths=0.0, c='k')
+ax.scatter(-uu,-vv, s=.5, rasterized=True, linewidths=0.0, c='k')
+ax.set_xlabel(r"$u$ [k$\lambda$]")
+ax.set_ylabel(r"$v$ [k$\lambda$]")
+ax.set_title(r'$U$, $V$ Visibilities')
+
+# As you can see, there are very few visibilities > 7,000 ($k\lambda$), and a very dense region of visibilities between -2000 and 2000 ($k\lambda$). This indicates outliers at the higher frequencies while the bulk of our data stems from these lower frequencies.
```
iancze, line 119: I wouldn't call the visibilities at long baselines outliers. The term [outlier](https://www.itl.nist.gov/div898/handbook/prc/section1/prc16.htm) has a specific statistical connotation that usually implies something with the data being suspect. Though the baseline lengths of these visibilities are indeed larger than the rest, the data is still quite valid (and useful). Instead of outlier you could just say there are several groups of long-baseline visibililities as well.

docs/tutorials/HD143006_Part_1.py
```diff
@@ -0,0 +1,211 @@
...
+ax[1].set_title('MPoL Dirty Image')
+ax[1].set_xlim(left=.75, right=-.75)
+ax[1].set_ylim(bottom=-.75, top=.75)
+ax[1].set_xlabel(r"$\Delta \alpha \cos \delta$ [${}^{\prime\prime}$]")
+ax[1].set_ylabel(r"$\Delta \delta$ [${}^{\prime\prime}$]")
+# fig.set_figwidth(14)
+fig.set_figheight(10)
+plt.tight_layout()
+
+# As you can see there are many similarities between the diagnostic dirty image and the image produced by the DSHARP survey using the CLEAN algorithm ([Andrews et al. 2018](https://ui.adsabs.harvard.edu/abs/2018ApJ...869L..41A/abstract)). While the dirty image is more noisy, it still maintains many distinct features present in the CLEAN image. In the next part of the HD143006 tutorial, we will be cleaning the diagnostic dirty image using RML through Neural Networks, Optimization, and Cross Validation with help from [PyTorch](pytorch.org).
```
iancze, line 210: This paragraph could use an edit. For example, what are the technical differences between the dirty image and the CLEAN image? What do you mean by "distinct features?" 

RML methods are not cleaning methods (those are CLEAN methods), so I wouldn't say that we're cleaning the dirty image using RML. RML and MPoL can take advantage of neural networks, but right now the MPoL package is not a neural network itself.

iancze (changes requested):
This is a very nice draft for the tutorial. I have a few suggestions in places to tighten up the text with grammatical and technical edits to be clear about what you are discussing.

docs/tutorials/HD143006_Part_1.py
```diff
@@ -0,0 +1,212 @@
...
+# 
+# Regularized Maximum Likelihood (RML) imaging is a forward modeling methodology. We predict an image then represent it as an array of pixels. This is brought to the visibility domain through a forward Fourier transform. This is then compared to the measured data to ensure it is a viable predicted image. Due to information loss of the true image when obtaining the measured data, several predicted images- including incorrect ones- will match. To get to our best predicted image, we make new predictions by choosing the most likely (Maximum Likelihood) configuration and favoring specified criteria (Regularized). These criteria or regularizers, are chosen by the user. Some examples of favored criteria are smoothness and sparsity. The likeliness and how well a predicted image meet a certain criterion is mathematically represented in a loss function that contains hyperparameters used to weight data and regularizers. We minimize this loss function by performing a gradient descent, in which we adjust the hyperparameters. When the loss function is minimized, our predicted image is at its best version to fit the collected data and follow our specified criterion.
```
hgrzy, line 29: After some discussion we determined that this may not be correct in that hyperparameters are set by the user (unless using something such as Ray Tune) and parameters are what are adjusted during gradient descent. Though I think there is still some grey area with these parameters. Would it be pixel size and/or Jy/arcsec^2 within each section of the cube that is changed with the gradient descent? If it is, how would this not result in the gradient descent producing our "true image" that we are comparing it with? Would it be because of the loss function not only looking at a comparison of the data from the true image but also taking into account what is most likely with the option to further analyze with regularizers?

iancze: This is a good question, and the answer is kind of subtle and kind of meta. The short answer is that yes, pixel size and number of pixels are technically parameters of the model, just like the way the intensity values of the pixels are parameters. But, in general, we'll want to leave pixel size and number of pixels constant during the fit.

The long answer is that this boils down to how one represents information (like an image). By choosing some pixel sampling rate, we're essentially asserting that the image is "band-limited" to those spatial frequencies. More information [here](https://en.wikipedia.org/wiki/Nyquist%E2%80%93Shannon_sampling_theorem).

hgrzy: @iancze Thank you! That helps me understand why we don't get the true image. I'm still a little unclear as to the adjustments made in the gradient descent. Above we state during the gradient descent the hyperparameters are what are being changed- or in the wording of Brianna's paper the coefficients corresponding to each regularizer. After reading Brianna's paper I'm wondering if it would instead be the pixel values usually represented with units of Jy arcsec^-2  that are being changed.

iancze: Perhaps I'm not quite following what you mean by "true image" in this context. In the case of HD 143006, we don't know what truth is, since this is real data! If we were to make a mock simulated dataset from a "true" sky image, like in the case of the ALMA logo, then we can go back and reference truth in that case.

The term "hyperparameters" can be thrown around a little loosely sometimes. First, hyperparameters *are* technically parameters, but the reason why they are "hyper" is because they are one level meta than the parameters under investigation in the base scenario. In the example of an RML fit, usually we would call the pixel value intensities (in Jy arcsec^2, or any unit, really) the parameters of the model. If we were doing a simple line fit to some x,y data, for example, we would call the slope and intercept the parameters of the model. It's these base level parameters that are what's being optimized by gradient descent.

We would typically call things like "regularization strength" or the lambda terms hyperparameters. For any given optimization run, we'll usually run with these parameters fixed. But there is the question of whether the hyperparameters could be varied to produce a "better" image (in the case of real data, better as measured by cross-validaton scroe). So this is where things like hyperparameter tuning (with Ray Tune) come in. There's a "meta-level" to the optimization, in that we're trying optimization with a bunch of different hyperparameter settings.

We could also treat the pixel sizes and number of pixels like hyperparameters, since they will have a (hopefully small) effect on the image that results. For now we're just ignoring this effect in the hyperparameter training to save computational effort.

hgrzy: I think you covered my questions and provided some helpful information! In a nutshell, I was thinking change "We minimize this loss function by performing a gradient descent, in which we adjust the hyperparameters," to "We minimize this loss function by performing a gradient descent, in which we adjust the pixel value intensities. Within this optimization run, hyperparameters are usually held fixed, but can be tuned between runs to produce a better image," though I wasn't sure if this was correct. However this conversation cleared it up! Thank you for your help @iancze !

iancze: Right, that's correct! The hyperparameters will not typically be changed during the optimization run or by gradient descent. I'm working on a few changes to the developer documentation + small updates to docs, so I can make this change on that branch, which I'll push and merge in a second.

hgrzy: Great! Thanks!

PR #78, opened 2021-06-14
[WIP] torchkbnufft and loose visibility connector

### LINKED ISSUE #17: "Loose" (Ungridded) Visibility DataConnector
The gridding process within the current [Gridder](https://github.com/MPoL-dev/MPoL/blob/main/src/mpol/gridding.py#L42) object makes assumptions about the data that, while fast to compute, obscure some potential calibration and processing subtleties. 

1) makes the assumption that the visibility function is constant across a u,v cell. This is technically something that could be improved within the GriddedDataset framework.
2) by collapsing all of the loose visibilities within a u,v cell to a single, averaged summary statistic (the gridded mean and gridded weight) on that cell, we lose the ability to visualize and assess differences between the loose visibilities. For example, we might want to examine if the loose visibilities (or residuals to them) have any trends with time of observation (suggesting a calibration error).

Provide a functionality to interpolate the dense Fourier grid to the loose visibilities, calculating a model visibility for every "loose" datapoint. This will allow us to implement things like spectral covariances (#18) and data self-calibration (#24).

Rather than bringing back the tried and true prolate spheroidal wavefunctions in earlier versions of MPoL and [DiskJockey](https://github.com/iancze/DiskJockey), I think it makes sense to try out [torchkbnufft](https://github.com/mmuckley/torchkbnufft) which uses Kaiser Bessel functions (similar to [FINUFFT](https://finufft.readthedocs.io/en/latest/index.html)). MRI and medical imaging in general has driven the development of these tools, and they're likely to be better than something that we would code from scratch.

**Describe alternatives you've considered**
It's possible that PyTorch conv2D layers could be adapted for the convolutional resampling, either from dense grid -> loose visibilities or even loose visibilities -> dense grid.

iancze:
When complete, will close #17 

Implementing "loose" visibility connector through the [torchkbnufft](https://github.com/mmuckley/torchkbnufft) package.


benjaminpope:
Out of curiosity - I just wanted to ask about your experience using the FINuFFT algorithm instead of an FFT.

In his [habilitation thesis](http://frantzmartinache.eu/static/share/hdr_no_papers.pdf) (in French - how's your French? It's really great), Frantz Martinache shows that if you want to go from the image to the _uv_ plane, doing an FFT and resampling gives _way_ worse model systematics than doing a matrix fourier transform (Figure 60 in that thesis).

I haven't used the NuFFT... how do we go here I wonder?

iancze:
I've heard great things about FINUFFT and would eventually like to use it. With the help of @dfm a while ago I think we had looked into how one might use it to calculate gradients (possibly in theano?), but I think he's probably solved those problems using his JAX bindings. I haven't seen PyTorch bindings, but if there were, I would go with that directly.

I'm not sure what kind of resampling is done in Frantz Martinache's thesis (my French is limited to bumbling about the patisserie!) but a naive resampling (e.g., bicubic) will indeed yield terrible results. When we've needed resampling (in [this](https://github.com/MPoL-dev/MPoL/blob/main/src/mpol/spheroidal_gridding.py) and in [other](https://github.com/iancze/DiskJockey/blob/main/src/gridding.jl) packages) we've used the "convolutional regridding" with spheroidal wavefunction approach. This has a history from the VLA and I think is shown to be sufficiently accurate (and we've [tested it](https://github.com/iancze/DiskJockey/blob/main/test/interpolation.jl), too). It works, but it's not the fastest. Do you know if "matrix Fourier transform" is the same thing that radio astronomers also call "direct Fourier transform," i.e., no "fast" involved?

After Alex Barnett (Flatiron) explained to me how FINUFFT works with Kaiser Bessel functions, I was persuaded that was probably a smarter and more efficient way to do things. In lieu of a PyTorch connection to FINUFFT, the Torchkbnufft package seems like a decent option. 

I think the main thing needed to move this forward is just for me to claw away a few days to figure out the API to torchkbnufft and connect up the wires--tricky with the start of the semester coming up. But last I remember working on this it seemed plausible that it should all hang together. We do actually need this functionality for another project we're involved with, so it's high priority.

benjaminpope:
Yeah I haven't used, but intend to try, [jax-finufft](https://github.com/dfm/jax-finufft) for `dLux` but haven't got around to it yet. 

Yeah in the Martinache work LDFT means local direct fourier transform, direct being with no 'fast' involved, as implemented in his `xara` package with [compute_dftm1](https://github.com/fmartinache/xara/blob/b82f7782ddcd744dc3ba975e3e39328d4c08b100/xara/core.py#L669) and [compute_dftm2](https://github.com/fmartinache/xara/blob/b82f7782ddcd744dc3ba975e3e39328d4c08b100/xara/core.py#L626).

PR #80, opened 2021-06-18
[WIP] HD 143006 part 2

iancze:
This is a draft pull request to monitor the changes between this branch and the main branch, as well as provide a vehicle for commentary and review.

To update the file contents, continue to push committed changes to the branch, i.e. `$git push origin HD143006_Part_2_branch`


docs/tutorials/HD143006_Part_2.py
```diff
@@ -0,0 +1,468 @@
...
+
+# + nbsphinx="hidden"
+# %run notebook_setup
+# -
+
+# # HD143006 Tutorial Part 2
+#
+# This tutorial is a continuation of the [HD143006 Part 1](HD143006_Part_1.html) tutorial. It covers the same content as the MPoL tutorials on [Optimization](optimization.html), [Initalizing with the Dirty Image](initializedirtyimage.html), and [Cross Validation](crossvalidation.html) but in a streamlined fashion and using real data. These other tutorials provide a more comprehensive breakdown of each step in this tutorial.
+#
+# This tutorial will be going through how to initialize the model, the imaging and optimization process, how to improve the process through crossvalidation, and how to analyze the results of our work with Tensorboard.
```
iancze, line 27: "how to improve the process" -> what do you mean here? What does cross validation help you understand?

docs/tutorials/HD143006_Part_2.py
```diff
@@ -0,0 +1,468 @@
...
+    uu=uu,
+    vv=vv,
+    weight=weight,
+    data_re=data.real,  # separating the real and imaginary values of our data
+    data_im=data.imag,
+)
+
+# -
+
+# We now have everything from the last tutorial loaded and can begin the process of optimization and cross validation to improve our image quality.
```
iancze, line 82: Cross validation on its own doesn't improve image quality, it just helps you understand how well the model generalizes to unseen data given your choice of hyperparameters. This can help you identify hyperparameter settings (i.e. regularizer strengths) that lead to better images, i.e., ones that generalize to new data better (and are likely a closer match to "truth").

docs/tutorials/HD143006_Part_2.py
```diff
@@ -0,0 +1,468 @@
...
+#
+# ### Getting the Dirty Image and Creating the Model
+#
+# First, we are going to get the dirty image from our gridder object. We will use the Briggs weighting scale and set `robust=0.0` here as these options lead to an already well optimized image (see [Part 1](HD143006_Part_1.html)).
```
iancze, line 86: I'm not sure what you mean by the options lead to an already well-optimized image. The dirty image is just a possible starting point for the optimization, it's not the optimized RML image itself.

docs/tutorials/HD143006_Part_2.py
```diff
@@ -0,0 +1,468 @@
...
+    model.forward()  # get the predicted model
+    sky_cube = model.icube.sky_cube
+
+    loss = loss_fn(sky_cube, dirty_image)  # calculate the loss
+
+    loss.backward()  # calculate gradients of parameters
+    optimizer.step()  # update the parameters
+# -
+
+# In this tutorial we will be using different methods of RML optimization so we have to save the model, letting us start from this clean starting point each time. Information on saving and loading models and the state_dict can be found [here](https://pytorch.org/tutorials/beginner/saving_loading_models.html).
```
iancze, line 128: What do you mean by different methods of RML optimization? It seems like you're using the same type of optimization, but under different regularization strength settings?

docs/tutorials/HD143006_Part_2.py
```diff
@@ -0,0 +1,468 @@
...
+            writer.add_scalar("loss", loss.item(), iteration)
+            writer.add_figure("image", log_figure(model, residuals), iteration)
+
+        loss.backward()  # calculate gradient of the parameters
+        optimizer.step()  # update the model parameters
+
+    return loss.item()
+
+
+# With our function done, all that is left is to set the variables, load the initialized model, set our hyperparameters, create our optimizer, and put the data in the correct format.
```
iancze, line 274: What do you mean here by "set the variables?" Which variables?

docs/tutorials/HD143006_Part_2.py
```diff
@@ -0,0 +1,468 @@
...
+    model, new_config, k_fold_datasets, MODEL_PATH, writer=cv_writer3
+)
+print(f"Cross Validation Score: {cv_score3}")
+# -
+
+# And here are the results in the Tensorboard. As we run through this optimizer using different hyperparameters in the config file we can analyze the different results to work towards a lower cross validation score. Also note that while it may seem strange that the loss values are not lowest with the lowest cross validation scores, different config files increase the weight of some loss functions (and others remove some completely) so the loss values are not perfectly equal representations across each configuration and run.
+
+cv_log_dir = logs_base_dir + "cv/"
+# #%tensorboard --logdir {cv_log_dir}
+## uncomment the above line when running to view Tensorboard
```
iancze, line 466: Would it be possible to show the trained model for the best CV score in the notebook directly? Since this isn't possible with tensorboard, I think you would need to add some more code to display the final image like in the other tutorials.

iancze (commented):
Very nice work! I left some comments throughout the document pointing out some vague statements that could be clarified (or deleted, in some instances). 

Could you take another pass or two through editing the text? Then I'll go ahead and update it relative to the main branch and merge it.

Thanks!

docs/tutorials/HD143006_Part_2.py
```diff
@@ -79,7 +79,7 @@
 
 # -
 
-# We now have everything from the last tutorial loaded and can begin the process of optimizing our model to set a better starting point.
+# We now have everything from the last tutorial loaded and can begin the process of optimization to improve our image quality and perform cross validation on our model to better determine hyperparameters.
```
RCF42, line 82: I think here we should probably just say we're starting the process of initializing the model and getting it closer to the dirty image since we don't get to cross validation until a while later in the tutorial. Also don't want to just restate what's happening the whole tutorial since  we already do that at the beginning.

hgrzy: Good point, I agree

hgrzy: I moved it to line 93 and omitted optimization and cross validation as they are in future sections

docs/tutorials/HD143006_Part_2.py
```diff
@@ -0,0 +1,481 @@
...
+)
+
+# -
+
+#
+# ### Getting the Dirty Image and Creating the Model
+#
+# First, we are going to get the dirty image from our Gridder object. We will use the Briggs weighting scale and set `robust=0.0` here as this option leads to a dirty image resembling the DSHARP CLEAN image (see [Part 1](HD143006_Part_1.html)).
+
+import torch
```
briannazawadzki, line 87: Fairly minor point, but I think you should move `import torch` and also `dirty_image = torch.tensor(img.copy())` to after you say "We now have everything from the last tutorial loaded and we can create the model." Since you didn't mention tensors/pytorch in part 1, I think it would read better if you introduced this aspect of it after saying we are caught up from part 1.

docs/tutorials/HD143006_Part_2.py
```diff
@@ -0,0 +1,481 @@
...
+
+# We now have everything from the last tutorial loaded and we can create the model.
+
+from mpol.precomposed import SimpleNet
+
+model = SimpleNet(coords=coords, nchan=gridder.nchan)
+
+# ## Initializing Model with the Dirty Image
+#
+# We now have our model and data, but before we set out trying to optimize the image we should create a better starting point for our future optimization loops. A good idea for the starting point is the dirty image, since it is already a maximum likelihood fit to the data. The problem with this is that the dirty image contains negative flux pixels, while we impose the requirement that our sources must have all positive flux values. Our solution then is to optimize the RML model to become as close to the dirty image as possible (while retaining image positivity).
```
briannazawadzki, line 101: A "better" starting point relative to what? A blank/uniform image? I think we could make this clearer by saying we need to pick a starting point, which could be basically anything, and the dirty image is a logical choice. Rather than the current language which sort of implies we are changing the existing starting point to the dirty image instead.

briannazawadzki: I also think we should be careful with the language used at the end of this line. The current text says "Our solution then is to optimize the RML model to become as close to the dirty image as possible (while retaining image positivity)" which is true in this case; we're only enforcing image positivity which is a very simply prior, but in most cases we're not trying to make an image as close to the dirty image as possible, we're trying to make a better image. I think it would be useful to rephrase either emphasizing that image positivity is the only prior being considered here, or that we are trying to make the best possible image given the dirty image and the need for positive pixel values.

docs/tutorials/HD143006_Part_2.py
```diff
@@ -0,0 +1,481 @@
...
+#
+
+
+# To optimize the RML model toward the dirty image, we will create our training loop using a [loss function](../api.html#module-mpol.losses) and an [optimizer](https://pytorch.org/docs/stable/optim.html#module-torch.optim). MPoL and PyTorch both contain many optimizers and loss functions, each one suiting different applications. Here we use PyTorch's [mean squared error function](https://pytorch.org/docs/stable/generated/torch.nn.MSELoss.html) between the RML model image pixel fluxes and the dirty image pixel fluxes.
```
briannazawadzki, line 105: I would cite the optimization tutorial for more info about the training loop.

docs/tutorials/HD143006_Part_2.py
```diff
@@ -0,0 +1,481 @@
...
+    test_score = np.sum(np.array(test_scores))
+
+    # adds cross validation score
+    if writer is not None:
+        writer.add_scalar("Cross Validation", test_score)
+
+    return test_score
+
+
+# Now, with our functions defined, we need to do the critical part of dividing our dataset into training and test datasets. There are many ways of going about this but here we are splitting it radially and azimuthally and removing chunks. This is visualized in the [Cross Validation tutorial](crossvalidation.html).
```
briannazawadzki, line 360: You already cite the cross validation tutorial so maybe this isn't necessary, but maybe mention that MPoL's Dartboard is an easy built-in way to get the polar coord grid, and maybe very briefly mention why we are choosing to do it this way.

docs/tutorials/HD143006_Part_2.py
```diff
@@ -0,0 +1,481 @@
...
+    )
+    plt.colorbar(im, ax=ax[1, 1])
+
+    return fig
+
+
+# With these set up, we can now make our training function (instead of a loop, we use a function here since the training loop will be ran multiple times with different configurations). The hyperparameters, such as `epochs` and `lambda_TV`, are contained under `config`. Most of them are used in the loss functions and can be read about [here](../api.html#module-mpol.losses).
+
+
+def train(model, dataset, optimizer, config, writer=None, logevery=50):
```
briannazawadzki, line 247: I'm a bit confused. Now you are introducing more priors than just the image positivity? I think you need to be more clear about what you're including, because earlier it seemed like you were saying that you were only using an image positivity prior.

trq5014: I agree with this, I just feel like I don't have the knowledge to explain each of the priors here. Adding the `lambda_sparsity` prior affects the sparsity of the image, the `lr` affects how aggressive the model learns, `epochs` determines the amount of iterations the model trains for, I can assume that `prior_intensity` affects the strangth of the `entropy` term, but I'm not quite sure what the prior `entropy` is. If you someone doesn't mind explaining this and confirming what I said about the other priors that would be very helpful.

briannazawadzki: Sure, I'll give my thoughts. I actually think it's okay if you don't fully explain all of the priors here, but I think it would be useful to explicitly state that now you are expanding beyond the initial optimization shown above, which only required image positivity.

The lambda terms affect the strength of the regularizing terms, but don't affect the form of the regularizer. They are basically just coefficients to weight each term. So sparsity, total variation (TV), and entropy are all different priors/regularizers that we add to the total loss term. I'm working on explaining each of them in my paper, but I haven't written everything yet. I think for the purposes of the tutorial it would be fine to just mention that you're now using multiple priors, and direct the reader [to the API](https://mpol-dev.github.io/MPoL/api.html#module-mpol.losses) for more information on these losses.

docs/tutorials/HD143006_Part_2.py
hgrzy: @briannazawadzki Are there cases when we don't initialize the model with the dirty image before beginning optimization for the best image? My understanding was that this section of the tutorial is initializing the model to the dirty image while retaining the image positivity prior, so we would want the dirty image to closely resemble the model at this point. Then later we take this as the starting point and begin to perform the actual optimization such that the model fits the visibilities but also takes into account any other priors so the model resembles what we'd expect the true image to be.

hgrzy:
@iancze Ready for another round of review!

docs/large-tutorials/HD143006_part_2.py
```diff
@@ -0,0 +1,480 @@
...
+
+# Now lets initialize the model to the pseudo-dirty image, set our hyperparameters, and create our optimizer.
+
+model.load_state_dict(
+    torch.load("model.pt")
+)  # load our initialized model from the previous section
+
+# Here is where we define the hyperparameters under the `config` dictionary. Most of these hyperparameters, such as `lambda_TV` and `entropy`, correspond to the scalar prefactors $\lambda$ as described in the [intro to RML](rml_intro.html).
+
+config = (
```
iancze, line 291: Where do these config values come from? The text doesn't provide any context to why they might be set to the values that they are.

docs/tutorials/HD143006_Part_2.py
briannazawadzki: @hgrzy Oops, I'm sorry, I must not have my GitHub notifications configured correctly. Yes, your understanding is correct. The dirty image is being used as the starting point for the optimization loop. But sure, there could be cases where you don't want to do that. For example, you might want to see if you can still converge on a reasonable solution using the same priors but starting with a uniform blank image. It would probably take more iterations to converge than using the dirty image, since presumably the dirty image is already in the ballpark of what the final result will be, while a blank image needs a lot more adjusting.

My comment was mainly just being nitpicky about language for clarity.

hgrzy: @briannazawadzki No worries! Ah that makes sense. Thanks for bringing up that good point.

iancze:
@trq5014 @hgrzy @RCF42 would it be possible to update the final CV example with some of the better-looking hyperparameter settings that you all found recently? It makes sense to show off our best image :)

After that, I think this looks ready to merge! Thanks for all of the hard work.

trq5014:
Yeah that can be done @iancze ! Do you want us to add the TSV prior in as well? Or should we stick with what we have?

iancze:
Might as well add in the TSV prior if it's needed to get the "best" image so far.

iancze:
Thanks for all the excellent work, everyone!

PR #85, opened 2021-06-28
Quick hyperlink update to docs

trq5014:
I noticed with the new folder update that some of the links that were hard coded (i.e [link](https://mpol-dev.github.io/MPoL/ci-tutorials/crossvalidation.html) rather than [link](crossvalidation.html)) and fixed this for future updates for the docs. I did not know how to create a general reference to .htmls built in other directories so references from HD143006 Tutorial Part 1 were updated, but will not be viable if the folder naming system is updated.


iancze:
Thank you!

hgrzy:
I think there is a discussion about accessing links in other directories here: https://stackoverflow.com/questions/10199233/can-sphinx-link-to-documents-that-are-not-located-in-directories-below-the-root

PR #93, opened 2021-08-19
[WIP] Stratified K fold CV

hgrzy:
Perform k fold cross validation where each k fold has the same ratio of low : high spatial frequency data points as the overall dataset.

Currently working on the for loop. I'm skeptical with the number of rows in the np array "pairs." While iterating through the while loop in the init of a stratkfold object it stops producing any flags that I've incorporated though jupyter notebook says it is still running the cell.

### DIFF
--- a/src/mpol/datasets.py
+++ b/src/mpol/datasets.py
@@ -445,3 +445,166 @@ class KFoldCrossValidatorGridded:
             return train, test
         else:
             raise StopIteration
+
+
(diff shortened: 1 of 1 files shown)


iancze:
Thank you for the contribution! Closing this PR now since the codebase and the design for CV has evolved significantly. Relevant code is reproduced in #182 for reference.

PR #103, opened 2022-10-29
WIP: PrimaryBeam correction

iancze:
WIP pull request to implement the PrimaryBeam correction for MPoL.


src/mpol/images.py
```diff
@@ -279,6 +280,120 @@ def to_FITS(self, fname="cube.fits", overwrite=False, header_kwargs=None):
...
+        nchan = len(np.at_least_1d(chan_freqs))
+        super().__init__()
+        _setup_coords(self, cell_size, npix, coords, nchan)
+        
+        self.telescope_to_radius = {"ALMA_12": 6.} # in meters
+        self.telescope_to_blocked_radius = {"ALMA_12": 0.375}
+        
+        assert (telescope in self.telescope_to_radius) and (telescope in self.telescope_to_blocked_radius)
+        
+        self.pb_mask = self.obscured_airy_disk(chan_freqs, telescope)
```
iancze, line 310: You might want to add a comment here noting that pb_mask is actually in packed cube format. I need to be a bit better about specifying these throughout the ImageCube too.

src/mpol/images.py
```diff
@@ -292,7 +407,7 @@ class FourierCube(nn.Module):
 
     def __init__(self, cell_size=None, npix=None, coords=None):
 
-        super().__init__()
+        super().__init__(cell_size)
```
iancze, line 410: Why does cell_size need to be passed here?

src/mpol/precomposed.py
```diff
@@ -50,6 +50,11 @@ def __init__(
         self.icube = images.ImageCube(
             coords=self.coords, nchan=self.nchan, passthrough=True
         )
+        
+        self.pbcube = images.PBCorrectedCube(
```
iancze, line 54: Rather than modifying SimpleNet, I think it might be better to subclass it as a new SimplePBNet and then make your modifications to that. I think that for learning purposes new users will want to start with literally the simplest thing possible, and the PB correction is a step up from that

iancze (commented):
Looking great so far!

iancze:
@kdesoto-psu I started a WIP pull request and added some comments for you here. If you continue to push to your main branch as you would normally do, this should continue to update and we can centralize comments here.

iancze:
Closing this PR #103 for now and will shortly re-open new PR to see if it provides better diff (not full duplication).

PR #113, opened 2023-01-19
add wavelength support to utils.convert_baselines

jeffjennings:
Slightly generalizes `utils.convert_baselines` to support a normalization wavelength argument (as an alternative to frequency).

### DIFF
--- a/src/mpol/utils.py
+++ b/src/mpol/utils.py
@@ -151,20 +151,27 @@ def fftspace(width, N):
     return xx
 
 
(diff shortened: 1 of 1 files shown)


jeffjennings:
There's some sphinx issue in building the docs, but that must be pre-existing, so requesting review.

iancze (approved):
Looks great, thanks. If you wanted to be even more thorough, you could add a test to test/utils_test.py

PR #114, opened 2023-01-20
prevent repeated var names in cross-val tutorial

### LINKED ISSUE #95: 'GriddedDataset' object is not callable
Sometimes the error message `'GriddedDataset' is not callable` appears. It resolves when I start a new job/reboot the kernel.

It can pop up at seemingly random times, but it frequently pops up after running an optimization loop, something like:
```
model.load_state_dict(torch.load("model.pt"))
optimizer = torch.optim.Adam(model.parameters(), lr=config["lr"])
writer = None
loss_val, loss_track = train(model, test4, optimizer, config, device=device, writer=writer)
vis = model.forward()
ground_test4 = utils.packed_cube_to_ground_cube(vis)
```
**if also followed by cross validation**, e.g.,
```
def cross_validate(config):
    test_scores = []

    for k_fold, (train_dset, test_dset) in enumerate(k_fold_datasets):
        rml = precomposed.SimpleNet(coords=coords, nchan=train_dset.nchan)
        optimizer = torch.optim.Adam(rml.parameters(), lr=config["lr"])
        train(rml, train_dset, optimizer, config)
        test_scores.append(test(rml, test_dset))

    test_score = np.sum(np.array(test_scores))
    return test_score, test_scores

cv_score, ind_cv_scores = cross_validate(config)
```
Unfortunately I haven't been able to pinpoint the issue in further detail. This error has also popped up at other times so it's not isolated to one script in particular. I know @trq5014 has also run into this before.

It would be great if we could find where this issue is coming from.

jeffjennings:
closes #95 

Just changing variable names in the cross-val tutorial (names `train` and `test` are used twice) to prevent redundant names and possible confusion for those taking code from the tutorial.

### DIFF
--- a/docs/ci-tutorials/crossvalidation.md
+++ b/docs/ci-tutorials/crossvalidation.md
@@ -185,11 +185,11 @@ The following plots visualize how we've split up the data. For each $K$-fold, we
 ```{code-cell}
 fig, ax = plt.subplots(nrows=k, ncols=3, figsize=(6, 10))
 
-for i, (train, test) in enumerate(k_fold_datasets):
+for i, (train_subset, test_subset) in enumerate(k_fold_datasets):
 
-    rtrain = connectors.GriddedResidualConnector(flayer, train)
+    rtrain = connectors.GriddedResidualConnector(flayer, train_subset)
     rtrain.forward()
-    rtest = connectors.GriddedResidualConnector(flayer, test)
+    rtest = connectors.GriddedResidualConnector(flayer, test_subset)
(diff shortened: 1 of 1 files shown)


jeffjennings:
I don't know if it matters that the docs tests are failing viz a viz building/deploying the docs to incorporate this change.

iancze (approved):
Thanks!

iancze:
Thanks! We'll merge now... the docs build error will eventually get fixed and then this will finally see the light of day.

PR #115, opened 2023-01-20
enforce KFoldCrossValidatorGridded to cpu

jeffjennings:
Moves all objects/operations in `datasets.KFoldCrossValidatorGridded` to the CPU, for ease of numpy usage.

### DIFF
--- a/src/mpol/datasets.py
+++ b/src/mpol/datasets.py
@@ -361,6 +361,7 @@ class KFoldCrossValidatorGridded:
         q_edges (1D numpy array): an array of radial bin edges to set the dartboard cells in :math:`[\mathrm{k}\lambda]`. If ``None``, defaults to 12 log-linearly radial bins stretching from 0 to the :math:`q_\mathrm{max}` represented by ``coords``.
         phi_edges (1D numpy array): an array of azimuthal bin edges to set the dartboard cells in [radians]. If ``None``, defaults to 8 equal-spaced azimuthal bins stretched from :math:`0` to :math:`\pi`.
         npseed (int): (optional) numpy random seed to use for the permutation, for reproducibility
+        verbose (int), default=0: level of verbosity of log messages
 
     Once initialized, iterate through the datasets like
 
@@ -380,7 +381,14 @@ class KFoldCrossValidatorGridded:
         q_edges=None,
         phi_edges=None,
         npseed=None,
+        verbose=0
     ):
+
+        # for ease of interface with numpy, enforce that griddedDataset tensors be on the CPU  
+        if griddedDataset.vis_gridded.is_cuda:
+            griddedDataset = griddedDataset.to('cpu')
+            if verbose > 0:
+                print('KFoldCrossValidatorGridded: moving griddedDataset to CPU')
 
         self.griddedDataset = griddedDataset
 


iancze (commented):
Have you tested this in a larger RML workflow? For example, within the same script or session (e.g., Jupyter notebook) I can imagine someone using the entire gridded dataset to do an inference loop (in which case, I think it needs to be on the GPU?) and also doing cross-validation.

jeffjennings:
> Have you tested this in a larger RML workflow? For example, within the same script or session (e.g., Jupyter notebook) I can imagine someone using the entire gridded dataset to do an inference loop (in which case, I think it needs to be on the GPU?) and also doing cross-validation.

Ah I thought you wanted the output of `KFoldCrossValidatorGridded` to be put on the cpu and later (optionally) moved by another routine to the GPU. Alright I've changed the class to take in a `device` in order to ensure it returns the datasets on the correct device. I also had to add a `device` arg to `FourierCube` to run a full workflow (including train/test/cross-val) on the gpu. I see a speedup of 5-10x for cross-val on my single gpu system. 

So I think everything that's worth putting on the gpu for speedup can now be moved there by passing these `device` args; it would be simple to do the same for `gridding.to_pytorch_dataset` and `precomposed.SimpleNet`, though I'm not sure there's any benefit to doing that, so I've left it out. Existing changes would benefit from gpu-specific tests, but I think that should be a separate PR.

iancze (commented):
Thanks, I think this makes sense, but I think it would be a good idea to develop and pass GPU-specific tests before we merge this into the main branch. Unless you've tested this locally in a few different workflow scenarios I'm worried we might run into confusing situations about where the data is.

They could either be part of this PR or as a separate PR. The tests could go within the normal test/datasets_test.py or we could create a new cross-validator test file. We could use environment variables or PyTorch's is_cuda to ascertain whether a GPU is available and whether we can run a test on it in addition to running on the CPU.

Then we could write a simple batch SLURM script for roar that gets a GPU and runs pytest.

jeffjennings:
Ok I looked at this again and have made some improvements - sorry they're after the review. 

1) I simplified the gpu/cpu back-and-forth in [c3d8380](https://github.com/MPoL-dev/MPoL/pull/115/commits/c3d8380f9bfbd0836c5c0c53217418a817a8597a) - I was making it too complicated before. 

2) I did decide to add the `device` arg to `gridding.to_pytorch_dataset` and `precomposed.SimpleNet`, in order to offer an integrated option to store the dataset and model on the GPU. The default behavior doesn't move them, but I think the option could be useful and saves the need to manually move them after they're created.  

No problem to include tests in this PR; in addition to covering cross-validator tests, they should probably also cover relevant tests for 2).

jeffjennings:
Ok I've now added gpu equivalents of existing tests for all the routines the PR modifies. Are there more gpu tests worth adding? 

Also would you mind writing the slurm script for roar?

src/mpol/fourier.py
```diff
@@ -19,10 +19,11 @@ class FourierCube(nn.Module):
         cell_size (float): the width of an image-plane pixel [arcseconds]
         npix (int): the number of pixels per image side
         coords (GridCoords): an object already instantiated from the GridCoords class. If providing this, cannot provide ``cell_size`` or ``npix``.
+        device (torch.device) : the desired device of the image cube. If ``None``, defalts to current device.
 
     """
 
-    def __init__(self, cell_size=None, npix=None, coords=None):
+    def __init__(self, cell_size=None, npix=None, coords=None, device=None):
```
iancze, line 26: Following @kadri-nizam 's discussion, I don't think it's necessary to specify `device` in this way. So long as the relevant aspects of the layer are properly registered as buffers or parameters, then if we had something like 

`fcube = Fourier(...)`

we could just do `fcube.to('cuda')`, which seems to follow PyTorch idioms.

iancze: I think similar comments apply to gridding and precomposed.

src/mpol/datasets.py
```diff
@@ -361,6 +361,7 @@ class KFoldCrossValidatorGridded:
         q_edges (1D numpy array): an array of radial bin edges to set the dartboard cells in :math:`[\mathrm{k}\lambda]`. If ``None``, defaults to 12 log-linearly radial bins stretching from 0 to the :math:`q_\mathrm{max}` represented by ``coords``.
         phi_edges (1D numpy array): an array of azimuthal bin edges to set the dartboard cells in [radians]. If ``None``, defaults to 8 equal-spaced azimuthal bins stretched from :math:`0` to :math:`\pi`.
         npseed (int): (optional) numpy random seed to use for the permutation, for reproducibility
+        device (torch.device) : the desired device of the output. If ``None``, defalts to current device.
```
iancze, line 364: I think the PyTorch datasets model is slightly different than nn.module, but a similar concept applies as the comments to the other modules. `UVDataset` is inherited from this class, so should be able to do this. `GriddedDataset` is not, however, so that might require a larger redesign.

iancze: #126 relevant to any Dataset refactoring, including KFoldCrossValidator inheriting from Dataset

jeffjennings: removed device arg

src/mpol/fourier.py
```diff
@@ -61,6 +63,9 @@ def forward(self, cube):
         # See MPoL documentation and/or TMS Eqn A8.18 for more information.
         self.vis = self.coords.cell_size**2 * torch.fft.fftn(cube, dim=(1, 2))
 
+        self.register_buffer("vis",
```
iancze, line 66: I don't think this extra line needs to be here.

src/mpol/fourier.py
jeffjennings: Device calls have now been removed.

src/mpol/fourier.py
jeffjennings: Removed.

iancze (approved):
Thanks!

PR #117, opened 2023-01-24
get_optimal_image_properties utility

jeffjennings:
Adds a utility function that just calculates an optimal (i.e. Nyquist) image pixel size and number of pixels given a desired total image width and a baseline (or a baseline distribution, in which case user can specify a percentile of the distribution out to which the pixel size will Nyquist sample).

'Baseline' here is technically u, v or q, as the routine makes no assumption about projection/deprojection, as noted in the docstring.

### DIFF
--- a/src/mpol/utils.py
+++ b/src/mpol/utils.py
@@ -1,4 +1,5 @@
 import numpy as np
+import math
 import torch
 
 from .constants import arcsec, c_ms, cc, deg, kB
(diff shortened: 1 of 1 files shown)


iancze (commented):
This is a great idea to save some headache! 

The percentile is an interesting idea, but from the user's perspective aren't they more likely going to want a pixel size that yields a higher spatial frequency than exists in the data (i.e., uv grid extends beyond the data with some buffer)? I can't think of many situations where the user would want to use a percentile < 100, since this would be throwing away your long-baseline data.

It might be out of the scope of this pull request, but I've been meaning to add some sort of figure or text to https://mpol-dev.github.io/MPoL/units-and-conventions.html#discretized-representation that explains in great detail (with examples) how changing cell_size and npix changes the number and size of u, v grid cells.

jeffjennings:
Thanks for your thoughts! I added a more verbose docstring in ef5ea47 to make it clear that a single baseline value can be passed in for the `q` arg to get the image properties (and ignore the `percentile` arg). 

Routinely with ALMA data, since the longest baseline regions are SNR < 2 and eventually < 1 (primarily due to sparse sampling), a model without a well-selected noise prior that closely fits those data can easily introduce high frequency artifacts into the visibility fit and image. With a noise prior, percentile=100 thus typically yields a large buffer in effect, as, e.g. with many dsharp datasets, the data extend to ~8-10 Mlambda, but are SNR<2 by ~4-6 Mlambda. 

A couple plots on the scaling of grid cell properties with cell_size and npix would be helpful, I'll create an issue just to document it. Though I think it's best as another PR. With your nufft implementation in place (thanks!), I'm moving to more of an 'ungridded' pipeline.

iancze:
Ah, I see, that makes good sense. Thanks!

PR #118, opened 2023-01-24
Check baseline units

jeffjennings:
Adds `utils.check_baselines`, which runs a couple sanity checks to see whether input baselines are in units of [\lambda], rather than [k\lambda]. Not foolproof, but can catch this mistake a lot of the time. 

Doesn't check whether baselines could be in [m], as the expected range of baselines in [m] for ALMA (~10 - 2e4) is similar to expected range in [k\lambda] (~1 - 1e4). Maybe you have a simple idea for such a check?

### DIFF
--- a/src/mpol/utils.py
+++ b/src/mpol/utils.py
@@ -179,10 +179,10 @@ def check_baselines(q):
                 "[\lambda].".format(max(q), max_feasible_q)) 
 
     if min(q) > min_feasible_q * 1e3:
-        raise Warning("Minimum baseline of {:.1e} is large for units of "
-                    "[\klambda]. Baselines must be in units of "
+        raise Warning("Minimum baseline of {:.1e} is large for expected "
+                    "minimum value of {:.1e}. Baselines must be in units of "
                     "[k\lambda], but it looks like they're in "
-                    "[\lambda].".format(min(q), min_feasible_q)) 
+                    "[\lambda].".format(min(q), min_feasible_q * 1e3)) 
 
 
 def convert_baselines(baselines, freq=None, wle=None):
--- a/src/mpol/utils.py
+++ b/src/mpol/utils.py
@@ -151,7 +151,7 @@ def fftspace(width, N):
     return xx
 
 
-def check_baselines(q):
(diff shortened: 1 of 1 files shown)


iancze (commented):
I think this is a good idea to include in a normal ALMA workflow. Thinking more generally, with aspirations for MPoL to be used on VLA data (or even optical/IR interferometric data), is there a way we could make this routine more foolproof? One might be to put `max_feasible_q` as a default argument, that could be changed depending on array. Another would be to allow baselines with astropy units that could be checked. Units get a bit messy since I don't think they play well with PyTorch, so we'd be adding them just to discard them later.

It's tricky that m and klambda have similar scales for ALMA. One possible solution is to have a user-facing routine that *expects* inputs in units of meters. This simplifies things for spectral line data because then u and v can be 1D nvis arrays even while data is a (nchan, nvis) array. Then the conversion to klambda occurs under the hood and is passed to the relevant MPoL routines. Note that this also could interface nicely to the NuFFT broadcasting https://mpol-dev.github.io/MPoL/ci-tutorials/loose-visibilities.html#the-mpol-fourier-nufft-object

jeffjennings:
Sure, I made `min_feasible_q` and `max_feasible_q` args. I think it's best to avoid unit conversions by keeping the routine native to code units. The function is designed as an internal check anyway. 

I would personally prefer the code intake either [m] or [lambda] and as you say convert to code units internally. But I think it's good to choose one unit and use it everywhere (including in user-facing), rather than having routines that handle both - to help prevent propagating the wrong unit in a workflow. I'd think that since people already use the code, the base unit of [klambda] shouldn't change. But it's up to you - either a user has to do the conversion to [klambda] themselves before supplying to mpol, or supply [m] directly but also supply frequencies/wavelengths.

jeffjennings:
Is there anything else needed for this? It's not as foolproof as it could be, but it will still catch a potentially common mistake of inputting wrong units [lambda].

iancze:
I think it looks good as is and will hopefully catch common mistakes with data units.

PR #120, opened 2023-01-25
Docs build/deploy tests

jeffjennings:
Fixes the failing docs build/deploy tests by pinning version of `sphinx_book_theme`.


jeffjennings:
It looks like the `docs` tests take ~6 min. To save some time, I'd suggest these don't need to be run on every push, and could be moved to their own github "action". E.g. in frank we run docs tests only when a review is requested on a PR that targets `main`, and again when the PR is closed:

```
name: docs
# Run this workflow when a review is requested on a PR that targets the main
# branch, or the PR is closed
on:
  pull_request:
    types: [review_requested, closed]
    branches: [main]

# Prevent multiple PRs from building/deploying the docs at the same time
concurrency:
  group: ${{ github.workflow }}

jobs:
  docs:
    <...>
```

jeffjennings:
Also is there a reason you're running every set of tests twice? `tests.yml` has `on: [push, pull_request]`, which is an AND not an OR operation. `on: push` will always run the tests.

iancze (commented):
Thanks for tracking this down!

iancze:
> Also is there a reason you're running every set of tests twice? `tests.yml` has `on: [push, pull_request]`, which will run tests twice on a push to an open PR. `on: push` would suffice.

Great suggestion. I guess this was my own inexperience getting set up with GitHub actions and it just stuck.

iancze:
> It looks like the `docs` tests take ~6 min. To save some time, I'd suggest these don't need to be run on every push, and could be moved to their own github "action". E.g. in frank we run docs tests only when a review is requested on a PR that targets `main`, and again when the PR is closed:

I think the reason I kept docs build in there is because of the way I was thinking about the tutorials as being "continuously integrated" as well and serving as their own set of tests. But I'm open to rearranging things if it makes the workflow (from the perspective of the developer) smoother!

jeffjennings:
That makes sense; I'd say the docs are still well-integrated if the build test runs on a review request and on a merge into main - it's up to you. My thought is that the tests take a long time, ~20 minutes, and cutting that down for the majority of pushes would be convenient. This saves 5 - 6 minutes, which is not a lot but something. 

To save another couple minutes, does the `dl_files.yml` workflow need to install package dependencies? I think that's done again in `tests.yml --> Install vanilla package`? I also don't know why the tests with coverage take 7-8 minutes to run - it's less than half that locally for me - but I don't have an idea how to improve that.

iancze:
> That makes sense; I'd say the docs are still well-integrated if the docs build test runs on a review request and on a merge into main - it's up to you. My thought is that the tests take a long time, ~20 minutes, and cutting that down for the majority of pushes would be convenient. This saves 5 - 6 minutes, not a lot but something.

Sure, that makes good sense to me. 

> To save another couple minutes, does the `dl_files.yml` workflow need to install package dependencies? I think that's done again in `tests.yml --> Install vanilla package`? I also don't know why the tests with coverage take 7-8 minutes to run - it's less than half that locally for me - but I don't have an idea how to improve that.

I think dl_files.yml does need to install dependencies (or, at least, python and astropy), since it uses it's caching behavior to download the logo_cube and HD143006 datasets needed to run tests and docs builds (in `docs/download_external_files.py`). My understanding of the way GitHub actions actions/cache@v3 works is that once these files are downloaded the first time, they should be cached. The cache key is built off of a hash of the contents of the `download_external_files.py` itself, so this should only need to be rerun when that file changes. And, every other GitHub actions should just use the cached files (I think). There might be a smarter way to structure this to cut down on the build times, though.

iancze:
Nice! It looks like you cut the test/build time in half for most cases. Ready to merge?

jeffjennings:
Yeah, feel free to merge if you're good with it! I ended up caching the python environment (for each python version) that references `setup.py`; it only caches the vanilla install, not any extras, but I think it's good enough.

iancze:
Awesome, thanks very much!

PR #124, opened 2023-01-30
Vis histogram 

### LINKED ISSUE #97: Support for plotting sensitivity-weighted baseline distributions
Commonly, we'll want to plot the u-v coverage of an observation. This is usually done by plotting marks at locations of the baselines. Sometimes, it would be useful to incorporate a representation of the sampling density and/or weighted sampling density. This would more readily demonstrate that there is not just a sampled/not-sampled aspect to the u-v coverage, but that there is a huge range of effective sensitivities for u-v "cells" that are sampled. And, at least for CLEAN-based imaging, that this can change based upon the gridding weighting options.

The gridding/averaging routines should provide the option to return the quantities needed to make these plots. I think this would be
* number of visibilities within each cell
* total "weight" of each cell = sum of all weight values in cell
* possibility to grid according to Cartesian or polar coordinates. This would require creating a separate gridding routine for the purpose of plotting, since this is probably different enough from the averaging routines for RML preparation.

**Additional context**
Some examples include [Petry et al](https://ui.adsabs.harvard.edu/abs/2020SPIE11449E..1DP/abstract) 

Screen Shot 2022-09-09 at 1 25 12 PM

jeffjennings:
- Adds `plot.py` to codebase

- Adds `vis_histogram`, to generate a figure of 1d and 2d histograms of (u,v) coverage 
     * see examples below

- closes #97
    * The only thing `vis_histogram` doesn't do relative to that issue is give an option for a cartesian 2d histogram - but I don't think that has much use in addition to the polar plot, as this routine is meant for data diagnostics, not code diagnostics.

Notes: 
* `vis_histogram` doesn't use `datasets.Dartboard` because I found it best to just allow `\phi` to take the range [-\pi, \pi] for the polar plot (as datasets may not be perfectly symmetric), and I added a `weight` option for the histograms - so creating a `Dartboard` instance and calling `get_polar_histogram` in this case would amount to having to alter both just to call `np.histogram2d`.

* no tests added yet; I'd like to wait until the fit pipeline is ready and use a single test call to generate all figures in `plot.py` via the pipeline, like 
 ```
def test_pipeline_figure_generation():
    """Check the full fit pipeline when producing all figures"""
    _run_pipeline(make_figs=True)
```


jeffjennings:
Example figures generated: 

- Standard histograms
vis_histogram

- Weighted histograms
vis_histogram_weighted

- Visibilities overplotted
vis_histogram_data

iancze (approved):
Thanks, this is so useful!

iancze (approved):
Awesome! Looks fantastic. Could you also add the `plot.py` automodule to the API.rst so that it gets built with the docs? Longer term we should revise the [Gridding and Diagnostic tutorial](https://mpol-dev.github.io/MPoL/ci-tutorials/gridder.html) to incorporate outputs from your `vis_histogram` technique.

jeffjennings:
Great, thanks! I added it to the API and opened #142  for plotting in the docs.

PR #125, opened 2023-01-31
Loss function convergence

jeffjennings:
A simple function to test whether the loss function (or any function generically) has converged as it iterates.

### DIFF
--- a/src/mpol/losses.py
+++ b/src/mpol/losses.py
@@ -380,3 +380,24 @@ def TSV(sky_cube):
     loss = torch.sum(diff_ll**2 + diff_mm**2)
 
     return loss
+
+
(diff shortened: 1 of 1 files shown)


iancze (commented):
This looks like a good start! Can you provide an example of how this might be used in practice? 

E.g., how would the [loop](https://mpol-dev.github.io/MPoL/ci-tutorials/optimization.html#iterating-the-training-loop) change to take advantage of this tracker?

This could be either as a docstring to this method or in a test.

jeffjennings:
I was going to use this function in multiple places, but I've been reorganizing local changes and am down to just one call, so I'm going to rename this branch (closing this PR), rename and re-spec the branch. I'm moving this function into a training class.

PR #128, opened 2023-02-01
Fit runner

jeffjennings:
Adds a full fit pipeline, enabling the ability to run mpol from the command line

TODO : 
- [ ] update json files
- [ ] get logger working
- [ ] test full pipeline
- [ ] test figure generation
- [ ] test both CPU and GPU support


iancze:
Closing as out of scope for the v0.3 core library, but may make sense as a pipeline example or other package, per our conversations. Thanks!

PR #127, opened 2023-02-01
Train/test class implementation

jeffjennings:
Creates classes for training/testing.

TODO:
- [x] add tests of train: with, without loss_lambda_guess (ensure lambda values being updated)
- [x] add tests of test

For reference, #129 adds cross-val class.

Separate PR should add function for `train_diagnostics` figure in `plot.py`.

PR #129, opened 2023-02-02
Cross-val class implementation

jeffjennings:
Creates class for cross-val loop.

This PR should only be merged into `main` after #127.

TODO:
- [x] add test of full cross-val loop with `run_crossval`; of `split_dataset` only


iancze (approved):
This looks really nice. We just talked about the device arguments, so once these are there I think this looks good to merge.

jeffjennings:
Ok do you mind looking over the changes since your review/comment and if all looks good, merge? Thanks!

src/mpol/crossval.py
```diff
@@ -0,0 +1,182 @@
...
+        # create a radial and azimuthal partition for the dataset
+        dartboard = Dartboard(coords=self._coords)
+
+        # use 'dartboard' to split full dataset into train/test subsets
+        subsets = KFoldCrossValidatorGridded(dataset, k=self._kfolds,
+                                        dartboard=dartboard,
+                                        npseed=self._seed)
+
+        if hasattr(self._device,'type') and self._device.type == 'cuda': # TODO: confirm which objects need to be passed to gpu
+            test_train_datasets = [(train.to(self._device), test.to(self._device)) for (train, test) in subsets]
```
iancze, line 116: I think this will work OK for most things we're working on now, but I think the eventual model that we should adopt is to send one dataset to the GPU at a time, do the training for that, send the results back tothe CPU, and then repeat for the next dataset subset. Because `test_train_datasets` is now a list that of a bunch of K copies of the dataset, there's a good chance that it won't fit on GPU memory all at once especially if the original dataset is large.

iancze (approved):
Looks good to me and ready to merge, with the possible exception of sending data to the GPU all at once. You might want to try this with a real ALMA dataset to see if you run into any memory issues as is. Thanks!

jeffjennings:
Cool, the current code works with a couple datasets I tried, but it's a good point - memory profiling of a full fit pipeline (in both ram and vram) would be very helpful too. I'll note this in an issue.

PR #130, opened 2023-02-03
tests.yml: move linter after mpol install

jeffjennings:
bugfix: changes order of `tests.yml` to install mpol (and test dependencies) before running linter


iancze (approved):
Good catch!

Issue #100, opened 2022-10-05
Hermitian pairs of visibilities for imaging and fitting


iancze:
It would be worthwhile reviewing/auditing how we treat raw visibilities and their Hermitian pairs through the fitting and imaging processes.

We know that Hermitian pairs must be included for any routine that uses an iFFT to synthesize an image, otherwise the image will not turn out to be real.

However, it's less clear whether they should be included when we are using an image to forward model the visibilities and evaluating a likelihood on those, because there is the potential for "double counting" measurements. I don't think it makes a technical difference (since all measurements would be double counted) and would just be a scale factor adjustment relative to other regularizer settings. However, it would be good to come to a consensus on this and document it thoroughly.

iancze:
The chi_square and log_likelihood routines should not accept Hermitian pairs, this would be equivalent to double-counting the data. This doesn't change the minimum of the loss function surface, but does change the parameter uncertainties that would be inferred.

It doesn't matter whether this happens in the "normalized log likelihood" functions, since the normalization over data points makes the calculation the same as whether it has Hermitian pairs or not.

jeffjennings:
Note to potentially come back to: in #132 , the `RandomCellSplitGridded` class takes in a `griddedDataset`, which contains Hermitian pairs, but the selection/splitting of data in `RandomCellSplitGridded` doesn't maintain that Hermitian property.

iancze:
I've come to the conclusion that it would be best that any routine that is evaluating a likelihood function (or a loss function) should not be involved with Hermitian pairs, whether they are "loose" or whether they are gridded. 

If Hermitian pairs were to be included in the likelihood function calculation, then, in order for the calculation to be accurate, the simple \chi^2 sum of data - model would need to be replaced with a multi-variate Gaussian likelihood that included covariances between data points. There would be a *perfect* covariance between each point and its Hermitian pair.

The one case where Hermitian pairs *do* need to be included is in any routine that is creating an image using the *inverse* FFT. This applies to all of the routines involved in making dirty images. It does not apply to any forward-modeling routine involved in RML, like [FourierCube](https://github.com/MPoL-dev/MPoL/blob/main/src/mpol/fourier.py#L15) or the [NuFFT](https://github.com/MPoL-dev/MPoL/blob/main/src/mpol/fourier.py#L186).

iancze:
[...]

PR #132, opened 2023-02-03
Cross-val splitter

jeffjennings:
- Moves `datasets.KFoldCrossValidatorGridded` to `crossval.DartboardSplitGridded`, updates imports of this throughout tests and tutorials. I don't like renaming a user-facing function, but I think with this PR's new splitting class (`crossval.RandomCellSplitGridded`) and new class that performs cross-validation (`crossval.CrossValidate`), the name `KFoldCrossValidatorGridded` is confusing (e.g. it's not doing the cross-val, just splitting).
    * moves tests using `crossval.DartboardSplitGridded` into `crossval_test`
    
- Adds another dataset splitting class `crossval.RandomCellSplitGridded`, which draws random cells from the gridded vis to build the (train, test) sets, returning an iterator of `GriddedDataset` tuples, consistent with `crossval.DartboardSplitGridded`.
    * Cells in the top 1% of gridded weight are included in every train set
    * The split sets are even in number:
```
 for (train, test) in DartboardSplitGridded:
     print(train.shape, test.shape)

              torch.Size([19580]) torch.Size([5798])
              torch.Size([19492]) torch.Size([5886])
              torch.Size([20734]) torch.Size([4644])
              torch.Size([21266]) torch.Size([4112])
              torch.Size([20440]) torch.Size([4938])

 for (train, test) in RandomCellSplitGridded:
     print(train.shape, test.shape)

              torch.Size([20353]) torch.Size([5025])
              torch.Size([20353]) torch.Size([5025])
              torch.Size([20353]) torch.Size([5025])
              torch.Size([20353]) torch.Size([5025])
              torch.Size([20354]) torch.Size([5024])
```

- Updates `crossval.CrossValidate` workflow to use these classes to split the data (`crossval.CrossValidate.split_dataset`), within a k-fold cross-val loop (`crossval.CrossValidate.run_crossval`)
    * generates one train and one test `GriddedDataset` at a time in the loop to keep memory footprint reasonable

- Haven't confirmed GPU compatibility yet (will address in #128 once I include calls to `crossval.CrossValidate` in the fit pipeline)

- Haven't added a stratified k-fold splitter - let's see how `RandomCellSplitGridded` does


jeffjennings:
Testing whether review request triggers `docs_build` workflow once or twice...

jeffjennings:
...just once!

briannazawadzki:
I agree with the renaming; I think this is moving in a good direction and needed to be done eventually. Is the call to import `crossval.RandomCellSplitGridded` a placeholder for now? I only see the `DartboardSplitGridded` class. I'm guessing it is, given the todo note inside the random cell part of `CrossValidator`?

jeffjennings:
> I agree with the renaming; I think this is moving in a good direction and needed to be done eventually. Is the call to import `crossval.RandomCellSplitGridded` a placeholder for now? I only see the `DartboardSplitGridded` class. I'm guessing it is, given the todo note inside the random cell part of `CrossValidator`?

Great! That's right, I haven't pushed most local changes yet, but will shortly.

jeffjennings:
Added note in #100 about changes here.

PR #137, opened 2023-02-05
_setup_coord method replaced with classmethods

kadri-nizam:
Classes that either take in a `coords: GridCoords` object or a `cell_size: float` and `npix: int` values used the `_setup_coord` method from coordinates.py. These classes have been edited to each have their own class methods for different ways of instantiating itself. 

The new change is such that the `__init__` method accepts only `coord: GridCoord` objects. It is still possible to initialize from `cell_size` and `npix` values via: `ClassName.from_image_properties(cell_size, npix, **kwargs)`

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -54,7 +54,7 @@ EXTRA_REQUIRES = {
 }
 
 EXTRA_REQUIRES["dev"] = (
-    EXTRA_REQUIRES["test"] + EXTRA_REQUIRES["docs"] + ["pylint", "black", "pre-commit"]
+    EXTRA_REQUIRES["test"] + EXTRA_REQUIRES["docs"] + ["pylint", "black", "pre-commit", "mypy"]
 )
 
 
@@ -80,9 +80,11 @@ setuptools.setup(
     packages=setuptools.find_packages("src"),
     package_dir={"": "src"},
     classifiers=[
-        "Programming Language :: Python :: 3",
+        "Programming Language :: Python :: 3.8",
+        "Programming Language :: Python :: 3.9",
+        "Programming Language :: Python :: 3.10",
         "License :: OSI Approved :: MIT License",
         "Operating System :: OS Independent",
     ],
-    python_requires=">=3.6",
+    python_requires=">=3.8",
 )
--- a/src/mpol/constants.py
+++ b/src/mpol/constants.py
@@ -1,11 +1,18 @@
+from __future__ import annotations
+
+from typing import Final
+
+import astropy.constants as apy_const
 import numpy as np
-from astropy.constants import c, k_B
 
 # convert from arcseconds to radians
-arcsec = np.pi / (180.0 * 3600)  # [radians]  = 1/206265 radian/arcsec
+arcsec: Final[float] = np.pi / (180.0 * 3600)  # [radians]  = 1/206265 radian/arcsec
 
-deg = np.pi / 180  # [radians]
+deg: Final[float] = np.pi / 180  # [radians]
 
-kB = k_B.cgs.value  # [erg K^-1] Boltzmann constant
-cc = c.cgs.value # [cm s^-1]
-c_ms = c.value # [m s^-1]
+# Boltzmann constant
+kB: Final[float] = getattr(apy_const, "k_B").cgs.value  # [erg K^-1]
+
+# Light speed
+cc: Final[float] = getattr(apy_const, "c").cgs.value  # [cm s^-1]
+c_ms: Final[float] = getattr(apy_const, "c").value  # [m s^-1]
--- a/examples/HD143006/common_data.py
+++ b/examples/HD143006/common_data.py
@@ -1,18 +1,19 @@
+import matplotlib.pyplot as plt
 import numpy as np
 import torch
-from mpol import (
-    losses,
-    coordinates,
-    images,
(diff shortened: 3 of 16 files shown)


setup.py
```diff
@@ -54,7 +54,7 @@ def get_version(rel_path):
 }
 
 EXTRA_REQUIRES["dev"] = (
-    EXTRA_REQUIRES["test"] + EXTRA_REQUIRES["docs"] + ["pylint", "black", "pre-commit"]
+    EXTRA_REQUIRES["test"] + EXTRA_REQUIRES["docs"] + ["pylint", "black", "pre-commit", "mypy"]
```
iancze, line 57: Should this be included in 'test' instead of 'dev', since I'm assuming we'll want some sort of check including type hints to run as part of the test suite (say, on GitHub actions)?

src/mpol/constants.py
```diff
@@ -1,11 +1,18 @@
...
 
 # convert from arcseconds to radians
-arcsec = np.pi / (180.0 * 3600)  # [radians]  = 1/206265 radian/arcsec
+arcsec: Final[float] = np.pi / (180.0 * 3600)  # [radians]  = 1/206265 radian/arcsec
+
+deg: Final[float] = np.pi / 180  # [radians]
 
-deg = np.pi / 180  # [radians]
+# Boltzmann constant
+kB: Final[float] = getattr(apy_const, "k_B").cgs.value  # [erg K^-1]
```
iancze, line 14: What's the reason for writing this using `getattr` instead of just naming the constant? This seems more verbose and less intuitive.

iancze: Is it necessary for the `Final` type hint (I'm assuming this means `const` or something like that)?

iancze (changes requested):
Hi Kadri,

Thanks for putting together this big request! It's looking very nice. I have a few clarifying questions:

1. It looks like the continuous integration tests didn't run on this pull request (just the docs build). Do you know why that might be? Do they pass locally?
2. The `_setup_coords` replacement looks good to me in principle. There are a lot of places throughout the codebase and documentation, though. Have you checked that the documentation builds fine locally and all of the tutorials work? It seems like it ran OK on GitHub actions, but it's a little weird that the tests didn't run so maybe the docs build is suspect too. The main interplay I am concerned about is with the `nchan` arguments. If I recall correctly, for some classes, `nchan` is needed while for others it may not be needed (`FourierCube`) and for classes like `NuFFT`, `nchan` can alter which type of batching is done (though this is determined from the shape of `uu` and `vv` in that case.

Thanks,

Ian

iancze:
Ah, I see the reason about the test build. This is a PR from your forked repo, not from a branch on the main dev. It looks like they are passing fine on your fork?

jeffjennings:
> Ah, I see the reason about the test build. This is a PR from your forked repo, not from a branch on the main dev. It looks like they are passing fine on your fork?

I didn't think of that when I changed `tests.yml` to only `on: push` rather than `on: [push, pull_request]`. Although I don't get why `docs_build.yml` didn't run, as it has: 
```
on:
  pull_request:
    types: [review_requested, closed]
    branches: [main]
```

src/mpol/constants.py
kadri-nizam: mypy gives an error with the usual way because of the way astropy exports their constants. This is a temporary work around on my local machine that I forgot to change as I haven't gotten around to defining a mypy.ini file yet.

kadri-nizam:
> Ah, I see the reason about the test build. This is a PR from your forked repo, not from a branch on the main dev. It looks like they are passing fine on your fork?

Yeah, I was surprised when I didn't see the tests run on the pull request. The build is passing on my fork though. I also tested locally before sending the PR.

src/mpol/constants.py
kadri-nizam: Final will hint to mypy that the values shouldn't change. We can do away with it, though I would import constants.py as a module instead of importing the constants within it to signify to developers that the values shouldn't change. This way, it is always clear that, for instance, `const.kB` should not be altered.

setup.py
kadri-nizam: Yes. This should probably go into test eventually. While I'm implementing the changes, I thought I'd set it in the dev requirements.

kadri-nizam:
@iancze 

RE: `nchan`

The current logic in the function sets `nchan = 1` if `None` is given.
https://github.com/MPoL-dev/MPoL/blob/98c1521800e0143179a92fbe4fc642850d674c6d/src/mpol/coordinates.py#L208-L211

With the proposed changes, `nchan` has a default argument of 1 in pertinent `__init__` functions. This covers the `None` case. If the user specifies an input value, then `nchan` will take on the user's value. 

Classes that take on `nchan` values from other arrays will continue to do so as those classes' `__init__` function does not have an `nchan` argument. For example:
https://github.com/MPoL-dev/MPoL/blob/9a9b5ede1d82d5d6f6d34c9f44aebb1ca431b284/src/mpol/gridding.py#L165-L173

This means the classmethod also doesn't have the `nchan` argument keeping functionality the same for those classes.
https://github.com/MPoL-dev/MPoL/blob/9a9b5ede1d82d5d6f6d34c9f44aebb1ca431b284/src/mpol/gridding.py#L207-L219

PR #138, opened 2023-02-06
Trigger GitHub actions on PR from fork

jeffjennings:
When a PR is opened from a fork, the event `pull_request` won't trigger a github action workflow. Meanwhile, having both `pull_request` and `push` as events will cause the workflow to run twice on a push to a PR originating from within the repo (i.e. not from a fork). The event `pull_request_target` will run on a PR from a forked repo, and it will prevent running the tests twice otherwise. PR applies this change for `tests.yml` and `docs_build.yml`.

### DIFF
--- a/.github/workflows/docs_build.yml
+++ b/.github/workflows/docs_build.yml
@@ -6,6 +6,8 @@ on:
   pull_request:
     types: [review_requested, closed]
     branches: [main]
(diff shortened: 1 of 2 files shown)


iancze (approved):
Nice work finding the right configuration!

PR #139, opened 2023-02-06
pre-release.yml: move linter after mpol install

jeffjennings:
Same fix as #130 (which was for `tests.yml`), but there I didn't think to check that the workflow order should also change in `pre-release.yml`. 

Note: all CI tests are running twice until #138 is merged.


iancze:
Thanks for sorting out the workflow configs!

Issue #126, opened 2023-01-31
Dataset handling and refactoring


kadri-nizam:
**Is your feature request related to a problem or opportunity? Please describe.**
- GriddedDataset requires a device parameter input which can be cumbersome.
- `KFoldCrossValidator` might benefit from inheriting from PyTorch data set where it will be easier to segment the boolean indices as different slices of a PyTorch dataset
- For multi-execution block/multi-GPU workflows, we might want to explore using DataLoader 

**Describe the solution you'd like**
To stay idiomatic with PyTorch, we can implement a `.to(device)` method that encapsulates the process of transferring the required tensors to the correct device.

IC: Let's focus this issue specifically on the (potential) redesign of dataset-focused classes, like `GriddedDataset` or `UVDataset` and how inference or cross-validation loops will interact with them. Issue #154 will track progress related to the  `Gridder` and methods to *create* a `GriddedDataset` in the first place.

iancze:
#18 and #31 are relevant in the design for possible solutions here. Closed issues #17 and #32 may provide additional things to think about in this context.

iancze:
Our next step should be to break out the issues @kadri-nizam raised in the original post into more bite-sized issues. 

But before we do that, I wanted to comment that I think major redesigns to `GriddedDataset` and `Gridder` are necessary to address the Hermitian pairs issue raised in #100. To solve the Hermitian pairs issue, I think it makes sense to break apart the gridding->imaging and gridding->GriddedDataset capabilities currently contained in `Gridder`. Although these gridding functionalities are similar, I think trying to combine them all in a single class actually makes things more complicated than they need to be.

Originally, I wrote the `Gridder` class to mimic the "gridding" of individual visibility data points to a UV grid so that they could be imaged using the inverse FFT, creating the dirty image. This is very useful for visualization and debugging and I propose that we keep something like this (perhaps with a more descriptive name, that somehow indicates its ability to produce dirty images). As far as I can currently tell, all of the code pertaining to the gridding->imaging functionalities is straight numpy.

[...]

PR #141, opened 2023-02-06
Modifications to datasets.py

kadri-nizam:
- Type hints added to datasets.py
- Assertions converted to raised exceptions
- Some `__init__` functions' logic have been simplified by implementing guard clauses
- KFoldCrossValidator now has a classmethod that intializes itself from Dartboard `q_edges` and `phi_edges` parameter
- Updated and regenerated HD143006 documentation 
   - Documentation now uses `Gridder.from_image_properties` initializer


src/mpol/coordinates.py
```diff
@@ -131,7 +131,7 @@ def __init__(self, cell_size: float, npix: int) -> None:
         self.packed_q_centers_2D = np_fft.fftshift(self.sky_q_centers_2D)
         self.packed_phi_centers_2D = np_fft.fftshift(self.sky_phi_centers_2D)
 
-        self.q_max = (
+        self.q_max = float(
```
iancze, line 134: Just curious, what's the reason for the float call? Shouldn't this already be as a float type?

src/mpol/datasets.py
```diff
@@ -126,7 +155,7 @@ def to(self, device):
 
 
 # custom dataset loader
-class UVDataset(Dataset):
+class UVDataset(torch_ud.Dataset):
```
iancze, line 158: I think this class is stale and we'll want to eventually rewrite it, using some of the ideas put forth in #126. But for now it can come along for the ride.

iancze (approved):
Great! Thanks for implementing these changes and spreading the type hints throughout the package.

src/mpol/coordinates.py
kadri-nizam: Some of the numpy fft calls end up returning values that the linter can't infer its type from. The float call just makes it explicit for use with mypy later.

src/mpol/datasets.py
kadri-nizam: You are right -- it is stale. I searched for references in the code base but there were none. I thought I'd keep it there in case I missed it.

PR #145, opened 2023-02-09
GriddedDataset: torch warn fix

jeffjennings:
Resolves a warning raised when passing a tensor to `GriddedDataset.add_mask`

### DIFF
--- a/src/mpol/datasets.py
+++ b/src/mpol/datasets.py
@@ -101,10 +101,7 @@ class GriddedDataset:
             device (torch.device) : the desired device of the dataset. If ``None``, defalts to current device.
         """
 
-        if torch.is_tensor(mask):
-            new_2D_mask = mask.clone().detach().to(device)
-        else:
-            new_2D_mask = torch.tensor(mask, device=device)
+        new_2D_mask = torch.Tensor(mask).detach().to(device)
         new_3D_mask = torch.broadcast_to(new_2D_mask, self.mask.size())
 
         # update mask via an AND operation, meaning we will only keep visibilities that are
--- a/src/mpol/datasets.py
+++ b/src/mpol/datasets.py
@@ -101,7 +101,10 @@ class GriddedDataset:
             device (torch.device) : the desired device of the dataset. If ``None``, defalts to current device.
         """
 
-        new_2D_mask = torch.tensor(mask, device=device)
+        if torch.is_tensor(mask):
+            new_2D_mask = mask.clone().detach().to(device)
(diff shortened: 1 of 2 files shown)


jeffjennings:
I don't know why MyPy is failing in the test, unless it's searching through 
```
        if torch.is_tensor(mask):
            new_2D_mask = mask.clone().detach().to(device)
````
with a numpy array.

**Comment by kadri-nizam** [comment]
@jeffjennings 

This is a [known issue when typing with PyTorch](https://github.com/pytorch/pytorch/issues/32824). The recommended way for checking is to check the type explicitly: `if isinstance(mask, torch.Tensor)`. This is the recommended approach by the devs at PyTorch as can be seen from the `is_tensor` [documentation](https://pytorch.org/docs/stable/_modules/torch.html#is_tensor)

That being said, my suggestion is to simply skip the check since we have `mask.clone().detach()` in the proceeding line. Doing `new_2D_mask = torch.Tensor(mask).detach().to(device)` will perform the same thing for both Tensors and ArrayLike objects.

**Review by kadri-nizam — DISMISSED** [review]
@jeffjennings 

This is a [known issue when typing with PyTorch](https://github.com/pytorch/pytorch/issues/32824). The recommended way for checking is to check the type explicitly: `if isinstance(mask, torch.Tensor)`. This is the recommended approach by the devs at PyTorch as can be seen from the `is_tensor` [documentation](https://pytorch.org/docs/stable/_modules/torch.html#is_tensor)

That being said, my suggestion is to simply skip the check since we have `mask.clone().detach()` in the proceeding line. Doing `new_2D_mask = torch.Tensor(mask).detach().to(device)` will perform the same thing for both Tensors and ArrayLike objects.

File: src/mpol/datasets.py [hunk 1]
```diff
@@ -101,7 +101,10 @@ def add_mask(
             device (torch.device) : the desired device of the dataset. If ``None``, defalts to current device.
         """
 
-        new_2D_mask = torch.tensor(mask, device=device)
+        if torch.is_tensor(mask):
+            new_2D_mask = mask.clone().detach().to(device)
+        else:
+            new_2D_mask = torch.tensor(mask, device=device)
```
kadri-nizam, line 107: Can be simply replaced with: `new_2D_mask = torch.Tensor(mask).detach().to(device)`

iancze:
Thanks!

jeffjennings:
Ah ok I see. Thanks for explaining, Kadri!

Issue #143, opened 2023-02-07
Some docs tutorial cells broken 


jeffjennings:
**Describe the bug**
Recent changes to some code routines broke individual jupyter cells in some of the docs tutorials - just a few cases: 

https://mpol-dev.github.io/MPoL/ci-tutorials/fakedata.html

https://mpol-dev.github.io/MPoL/ci-tutorials/initializedirtyimage.html

https://mpol-dev.github.io/MPoL/ci-tutorials/gridder.html

**Additional context**
The `docs_build` workflow should catch this, so either it didn't run on the relevant merge commits, or it's not catching errors in the tutorial notebook cells.

jeffjennings:
This should ideally be fixed before the next release - adding it to the milestone.

iancze:
Thanks for flagging these. Strange that they weren't caught by the docs build, it's designed to catch these errors in the tutorials (and has in the past). @kadri-nizam did you run into this when you built your docs changes locally?

iancze:
I tried to build the docs locally and I now see the problem. Since I changed over to the MystNB system notebook build exceptions have become `WARNING`s instead of errors. I'll see if there's a build flag that will change this.

iancze:
Ok, #144 was just merged, which should now make docs builds error with notebook errors. So all broken docs PRs should also have a failing build. The build on `main` will also be broken now, too

kadri-nizam:
It seems that the failing builds are associated with the new `__init__` methods that I pushed in #137. I'll handle these errors.

kadri-nizam:
@iancze 

Issues due to change with the `__init__` function has been resolved. However intializedirtyimage.md is still throwing an error:
```
RuntimeError                              Traceback (most recent call last)
Cell In[16], line 1
----> 1 rml.load_state_dict(torch.load("dirty_image_model.pt"))
      2 rml.state_dict()  # the reloaded parameters of the model

File ~/Documents/GitHub/MPoL/.venv/lib/python3.8/site-packages/torch/nn/modules/module.py:1671, in Module.load_state_dict(self, state_dict, strict)
   1666         error_msgs.insert(
   1667             0, 'Missing key(s) in state_dict: {}. '.format(
   1668                 ', '.join('"{}"'.format(k) for k in missing_keys)))
   1670 if len(error_msgs) > 0:
-> 1671     raise RuntimeError('Error(s) in loading state_dict for {}:\n\t{}'.format(
[...]

PR #149, opened 2023-02-09
#137 Doc Errors

kadri-nizam:
Addressed issues mentioned in #143 that are related to #137.
 
I checked locally and the docs are still throwing an error which is not related to this PR. I'll document it in the aforementioned issue.


iancze (approved):
Looks great, thank you!

PR #156, opened 2023-02-18
Split averaging and imaging capabilities of Gridder into different classes.

### LINKED ISSUE #110: Residual Imager with Briggs Weighting
After #17 is implemented, it will be possible to make residual images using Briggs weighting. Currently, it is only possible to visualize residuals using uniform weighting, as the visibilities have been gridded. This isn't ideal, because it means that the residuals have very high thermal noise levels and therefore any structure is hard to recognize.

* Create an easy workflow that calculates loose residual visibilities and then uses the Dirty Imager to make a Briggs-weighted image.
* Write a tutorial demonstrating this functionality, or add to an existing tutorial

iancze:
This PR implements #154, by splitting the averaging capabilities of `Gridder` into `DataAverager` and the imaging capabilities of `Gridder` into `DirtyImager`. This should address the concerns in #100 . The use of `Gridder` was ubiquitous throughout the codebase, so this PR touches many files. 

For this PR to be merged, I think we need to *remove* `GriddedResidualConnector` and the `connectors.py` file entirely. This is because, with the output from DataAverager -> GriddedDataset no longer Hermitian, it doesn't make sense to have a direct imaging capability for the residuals, which is the main thing that `GriddedResidualConnector` provides. Instead, the imaging needs to go through something like #110 . It might make most sense to address this in a separate PR that can then be merged into this PR, rather than addressing this directly here.

Remaining items:
- [x] (Most likely in a separate PR) document the preferred way to image residuals and close #110 
- [x] (Most likely in a separate PR) remove the `GriddedResidualConnector` class entirely, merge that into `main`, and then update this branch.
- [x] Change doc strings of new `DataAverager` and `DirtyImager` classes
- [x] Modify tutorials to switch to `DataAverager` or `DirtyImager` where appropriate. Modify diagnostics tutorial to explain the difference between the classes.
- [x] Finalize any changes to tests and documentation such that everything passes.


iancze:
I think this is ready for review and possible merging. I think we might have a stale test error?

iancze:
Tests now passing and ready for a review.

docs/ci-tutorials/loose-visibilities.md
```diff
@@ -104,12 +104,12 @@ nufft = fourier.NuFFT(coords=coords, nchan=nchan, uu=uu_chan, vv=vv_chan)
 
 Now let's put the NuFFT aside for a moment while we initialize the {class}`mpol.gridding.Gridder` object and create an image for use in the forward model.
```
jeffjennings, line 105:
```suggestion
Now let's put the NuFFT aside for a moment while we initialize the {class}`mpol.gridding.DataAverager` object and create an image for use in the forward model.
```

src/mpol/crossval.py
```diff
@@ -155,63 +174,67 @@ def run_crossval(self, dataset):
...
-            self._cv_diagnostics = defaultdict(list) 
+            self._cv_diagnostics = defaultdict(list)
 
         split_iterator = self.split_dataset(dataset)
         for kk, (train_set, test_set) in enumerate(split_iterator):
             if self._verbose:
-                logging.info("\nCross-validation: k-fold {} of "
-                            "{}".format(kk, self._kfolds))
+                logging.info(
+                    "\nCross-validation: k-fold {} of " "{}".format(kk, self._kfolds)
```
jeffjennings, line 193:
```suggestion
                    "\nCross-validation: k-fold {} of {}".format(kk, self._kfolds)
```

src/mpol/gridding.py
```diff
@@ -140,19 +144,17 @@ def verify_no_hermitian_pairs(uu, vv, data, test_vis=5, test_channel=0):
...
 
-    If you want to take a quick look at the rough image plane representation of the visibilities, you can view the 'dirty image' and the point spread function or 'dirty beam' using the :func:`~mpol.gridding.Gridder.get_dirty_image` and :func:`~mpol.gridding.Gridder.get_dirty_beam` methods.
+    Subclasses will accept "loose" *ungridded* visibility data and store the arrays to the object as instance attributes. The input visibility data should be the set of visibilities over the full :math:`[-u,u]` and :math:`[-v,v]` domain, and should not contain Hermitian pairs (an error will be raised, if they are encountered).  The visibilities can be 1d for a single continuum channel, or 2d for image cube. If 1d, visibilities will be converted to 2d arrays of shape ``(1, nvis)``. Like the :class:`~mpol.images.ImageCube` class, after construction, GridderBase assumes that you are operating with a multi-channel set of visibilities. These routines will still work with single-channel 'continuum' visibilities, they will just have nchan = 1 in the first dimension of most products.
```
jeffjennings, line 155:
```suggestion
    Subclasses will accept "loose" *ungridded* visibility data and store the arrays to the object as instance attributes. The input visibility data should be the set of visibilities over the full :math:`[-u,u]` and :math:`[-v,v]` domain, and should not contain Hermitian pairs (an error will be raised, if they are encountered).  The visibilities can be 1d for a single continuum channel, or 2d for an image cube. If 1d, visibilities will be converted to 2d arrays of shape ``(1, nvis)``. Like the :class:`~mpol.images.ImageCube` class, after construction, GridderBase assumes that you are operating with a multi-channel set of visibilities. These routines will still work with single-channel 'continuum' visibilities, they will just have nchan = 1 in the first dimension of most products.
```

jeffjennings (approved):
Looks great, a few very small suggested changes.

iancze:
Thanks very much, @jeffjennings !

PR #160, opened 2023-02-21
Remove extraneous `.forward()` calls 

### LINKED ISSUE #151: Direct `.forward()` calls not recommended
I noticed in the code base that `model.forward()` is being called directly instead of `model()`. PyTorch devs recommend against the former since the `__call__` dunder method for the latter has [special hooks included in them](https://github.com/pytorch/pytorch/blob/36dfbb08f3dce0b082645e2499d8ddd7cde68fde/torch/nn/modules/module.py#L1494).

iancze:
This PR removes all of the extraneous `.forward()` calls that I could find in favor of calling the module directly. For example, `model()` versus `model.forward()`.

It should also fix a time-out error build with the docs.

Closes #151


jeffjennings (approved):
Straightforward - looks good to me!

iancze:
Thanks!

PR #165, opened 2023-02-22
Cross-val args, syntax

jeffjennings:
Alters some of the variables, properties of `CrossValidate` and `TrainTest` for ease of use and to store what's needed for diagnostics.

Todo: 
- [x] update tests


iancze (approved):
Looks good!

PR #169, opened 2023-02-28
split_diagnostics_fig: change presentation

jeffjennings:
Just changes the `split_diagnostics_fig` figure to show the test set in both columns so that it's easier to visualize where the full dataset is split: 
test_split_diag


iancze (approved):
Yea, I like this format better! Thanks :)

PR #171, opened 2023-03-04
Faster histograms

kadri-nizam:
I benchmarked a small part of the code base and found that the `np.histogram2d` call in `_sum_cell_values_channel` method is pretty slow. The method takes up almost 75% of the entire running time (22.2 s / 29.6 s). See the visualized profiling below with the relevant section highlighted in magenta.
image
Since our gridding process creates grids that are uniformly spaced, we can opt for libraries that specialize in histograms for uniform bin sizes. I added `fast-histogram` into project's dependency and substituted the histogram call. The method call now only takes 6.2% of the total running time (0.565 s / 9.12 s) resulting in a 3.25x speed up. Relevant section highlighted again in magenta below.
image

### DIFF
--- a/src/mpol/gridding.py
+++ b/src/mpol/gridding.py
@@ -6,10 +6,10 @@ from __future__ import annotations
 import warnings
 
 import numpy as np
+from fast_histogram import histogram as fast_hist
 
 from mpol.coordinates import GridCoords
-from mpol.exceptions import (DataError, ThresholdExceededError,
-                             WrongDimensionError)
+from mpol.exceptions import DataError, ThresholdExceededError, WrongDimensionError
 
 from .datasets import GriddedDataset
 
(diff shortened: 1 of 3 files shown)


src/mpol/gridding.py
```diff
@@ -246,15 +246,17 @@ def _sum_cell_values_channel(self, uu, vv, values=None):
             A 2D array of size ``(npix, npix)`` in ground format containing the summed cell quantities.
         """
 
-        result = np.histogram2d(
+        bin_range = self.coords.vis_ext[:2]
```
iancze, line 249: Could we make this code clearer to understand? For example, using `coords.v_bin_min`, and `coords.v_bin_max` and `coords.u_bin_min` etc..

iancze (approved):
Thanks, I think this is a nice improvement for a routine that does some heavy lifting (at least in the `DataAverager` and `DirtyImager` routines). It seems like the package is reasonably well maintained, otherwise I would be cautious about adding another core dependency outside of `numpy`.

src/mpol/gridding.py
kadri-nizam: Sure thing! Somehow it didn't occur to me that those attributes were available. I'll push the changes shortly.

kadri-nizam:
> otherwise I would be cautious about adding another core dependency outside of numpy.

I concur. I'll look into whether numpy has routines that are made specially for uniform bin sizes.

kadri-nizam:
It looks like the author of `fast-histogram` also contributed to making [numpy's histogram faster](https://github.com/numpy/numpy/pull/6100). They made their own version that's fully implemented in C without the bells and whistles that numpy does.

iancze:
Looks great, thanks! Given the fast-histogram author's commitment to various open source projects, I don't think it's a huge risk that that package will become outdated before MPoL does :) And, if it does, it should be easy enough to switch back to numpy-only.

PR #172, opened 2023-03-04
`fourier.py` type hinted

kadri-nizam:
The module is fully type hinted now. I also simplified certain logic, mostly where branching can be removed. The logic changes are in the NUFFT `__init__`, `forward`, and `_assemble_ktraj` method.

### DIFF
--- a/src/mpol/fourier.py
+++ b/src/mpol/fourier.py
@@ -239,54 +239,59 @@ class NuFFT(nn.Module):
 
     def __init__(
         self,
-        coords=None,
-        nchan=1,
-        uu=None,
-        vv=None,
-        sparse_matrices=True,
+        coords: GridCoords,
+        uu: NDArray[floating[Any]],
+        vv: NDArray[floating[Any]],
+        nchan: int = 1,
+        sparse_matrices: bool = True,
     ):
         super().__init__()
 
+        if not (same_uv := uu.ndim == 1 and vv.ndim == 1):
+            import warnings
+
+            warnings.warn(
+                "Provided uu and vv arrays are multi-dimensional, suggesting an intent to parallelize using the 'batch' dimension. This feature is not yet available in TorchKbNuFFT v1.4.0 with sparse matrix interpolation (sparse_matrices=True), therefore we are proceeding with table interpolation (sparse_matrices=False).",
+                category=RuntimeWarning,
+            )
+            sparse_matrices = False
+            self.interp_mat = None
+
         self.coords = coords
         self.nchan = nchan
+        self.sparse_matrices = sparse_matrices
+        self.same_uv = same_uv
 
         # initialize the non-uniform FFT object
         self.nufft_ob = torchkbnufft.KbNufft(
             im_size=(self.coords.npix, self.coords.npix)
         )
 
-        if (uu is not None) and (vv is not None):
-            self.register_buffer("k_traj", self._assemble_ktraj(uu, vv))
-        else:
-            raise ValueError("uu and vv are required arguments.")
-
-        self.sparse_matrices = sparse_matrices
+        self.register_buffer("k_traj", self._assemble_ktraj(uu, vv))
+        self.k_traj: torch.Tensor
 
         if self.sparse_matrices:
-            if self.same_uv:
-                # precompute the sparse interpolation matrices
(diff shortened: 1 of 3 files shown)


src/mpol/fourier.py
```diff
@@ -218,56 +243,63 @@ class NuFFT(nn.Module):
...
+        coords: GridCoords,
+        uu: NDArray[floating[Any]],
+        vv: NDArray[floating[Any]],
+        nchan: int = 1,
+        sparse_matrices: bool = True,
     ):
         super().__init__()
 
+        if not (same_uv := uu.ndim == 1 and vv.ndim == 1):
+            import warnings
```
iancze, line 255: It seems like this will raise a warning if the user specifies `sparse_matrices=False`?

iancze (commented):
Thanks, I think this contributes many helpful changes overall. The one place where I want to make sure I understand the changes better is the flow of logic in the NuFFT, since the original intent was to match this up with the way TorchKbNuFFT works as efficiently as possible. There are a few different branching possibilities (which should be covered in the tests), but I just want to make extra sure that these changes don't break that, since that hurt my brain to code (and is seemingly correctly, so far). Can you double check that the branching options of 
* single chan vs. multi-chan image cube
* single chan uu and vv (broadcast to multi chan image cube)
* single chan uu and vv (applied to single chan cube)
* multi-chan uu and vv (applied to multi-chan image cube)

are all handled and broadcast correctly using the batch/coil dimensions as needed?

(As you're no doubt aware) the shape considerations are laid out in the [docstring for NuFFT](https://mpol-dev.github.io/MPoL/api.html#mpol.fourier.NuFFT). I'd like to preserve the ability to use sparse matrices with multi-channel image cubes with `same_uv=False` in the future, does flow do that? 

Also, can you say a little more about what the protocol is doing?

src/mpol/fourier.py
kadri-nizam: Thank you for catching that! Looks like I did miss checking for sparsity there. I'll add that in.

kadri-nizam:
> ... are all handled and broadcast correctly using the batch/coil dimensions as needed?

I went through the logic again and believe it is the same as the previous implementation. I'm happy to pair-code with you over one of our hackathon sessions to ensure I didn't miss anything though. We can keep the PR open until then.

---

> I'd like to preserve the ability to use sparse matrices with multi-channel image cubes with same_uv=False in the future, does flow do that?

I based the refactoring based on the current implementation and I don't think the current implementation accounts for that case right now. If `sparse_matrices=True` and `same_uv=False`, the `else` clause on L253 gets executed which sets `sparse_matrices=False` on L261.
https://github.com/MPoL-dev/MPoL/blob/0cd0042c506cd683302d0dae3a1a0a667798ee0a/src/mpol/fourier.py#L244-L261
How would you like to proceed?

---

> Also, can you say a little more about what the protocol is doing?

Sure! Python's [Protocol](https://peps.python.org/pep-0544/) helps with structural duck-typing. In this case, the function:
```python
def get_vis_residuals(
    model: MPoLModel,
    ...
```
has a required parameter `model` which generally is a sub-class of `nn.Module`. However, we also expect it to have additional attributes that are related to MPoL classes (`GridCoords`, `ImageCube`, etc; the [`precomposed.SimpleNet`](https://github.com/MPoL-dev/MPoL/blob/0cd0042c506cd683302d0dae3a1a0a667798ee0a/src/mpol/precomposed.py#L8) model is an example). So with the `MPoLModel` protocol:
```python
class MPoLModel(Protocol):
    coords: mpol.coordinates.GridCoords
    nchan: int
    bcube: mpol.images.BaseCube
    icube: mpol.images.ImageCube
    fcube: mpol.fourier.FourierCube

    def forward(self):
        ...
```
Python linters will let the user know that any custom model they pass to the function **must** -- at a minimum -- implement the attributes as defined in the protocol *in addition to* a `forward` method to satisfy both requirements of MPoL and the `nn.Module`.

iancze:
> I went through the logic again and believe it is the same as the previous implementation. I'm happy to pair-code with you over one of our hackathon sessions to ensure I didn't miss anything though. We can keep the PR open until then.

Great, I think that will be the quickest route to make sure we're both on the same page.

The protocol stuff sounds nice, thanks.

PR #186, opened 2023-04-04
Redesigned GriddedDataset to inherit from `nn.Module`

### LINKED ISSUE #163: Redesign GriddedDataset with Pytorch idioms in mind
This is the second part of dataset redesign (first parts discussed in #126 and #162), focused specifically on the changes to `GriddedDataset`.

All of the same design challenges present in the `UVDataset` redesign (#162) also apply here. Additionally, we have the concern that we need to deal with masks.

One immediate problem with the current design of `GriddedDataset` is that it does not inherit from any Pytorch class. There is a workflow where it might make sense to attach the dataset as an instance attribute of some class (inherited from nn.module), and then use the `.to(device)` syntax to send everything, including the dataset, to the GPU. We tried to duck type this behavior with the `GriddedDataset.to` method, but I don't think this entirely meshes with this workflow. We would like to have seamless transfer/to from CPU/GPU using Pytorch idioms.

And, thinking further down the line, we would like this dataset object to play well with cross-validation. 

TODO: flesh this issue out with more details on constraints/desired features.

iancze:
Implemented the changes we discussed during the hackday: 
* `GriddedDataset` now inherits from `nn.Module`
* We made this decision (as opposed to a `torch.utils.data.Dataset` inheritance) because we could then write a `forward` method to do the indexing
* This had the added benefit that we could move the `index_vis` method inside the forward method of the new `GriddedDataset`
* Minor updates to tutorials
* Seems to work with GPU transfer

Closes #163

### DIFF
--- a/docs/large-tutorials/pyro.md
+++ b/docs/large-tutorials/pyro.md
@@ -248,7 +248,6 @@ import torch
 from torch import nn
 from mpol import geometry, gridding, images, fourier, utils
 from mpol.constants import deg
-from mpol.datasets import index_vis
 
 import pyro
 import pyro.distributions as dist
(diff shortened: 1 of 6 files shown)


iancze:
Tests pass locally on GPUs but I'm getting some mypy errors. Perhaps @kadri-nizam can chime in on what's needed here to resolve the errors?

PR #187, opened 2023-04-05
GridCoords constructor with `@properties` and `@cached_properties`

### LINKED ISSUE #152: GridCoords object has many instance attributes that could be made methods decorated with `property`
The `coordinates.GridCoords` class creates an object that stores the dimensions of the image plane grid and many other useful derived grids and quantities. The only real parameters of this class are `cell_size` and `npix`---every other quantity is derived.

Rather than instantiate these as real instance attributes in `__init__`, it would be more memory efficient to make all of these derived quantities their own methods that could be decorated with [`@property`](https://realpython.com/python-property/). This might even enable some additional error checking in some cases.

kadri-nizam:
Resolves #152

Repeated identical calculations for defining certain attributes are now done only once in the constructor and accessed via `@properties`. Non-identical attributes are set as `@cached_properties` and only computed once on first call.


src/mpol/coordinates.py
```diff
@@ -50,109 +51,208 @@ class GridCoords:
...
-        lmax = cell_size * (self.npix // 2 - 0.5)
-        lmin = -cell_size * (self.npix // 2 + 0.5)
-        self.img_ext = [lmax, lmin, lmin, lmax]  # arcsecs
+        # Image related
+        self._dimage = cell_size * const.arcsec  # [radians]
+        self._image_centers = self._dimage * (np.arange(npix) - npix // 2)
 
-        self.dl = cell_size * const.arcsec  # [radians]
-        self.dm = cell_size * const.arcsec  # [radians]
+        # Frequency related
```
iancze, line 70:
```suggestion
        # Spatial frequency related
```

src/mpol/coordinates.py
```diff
@@ -50,109 +51,208 @@ class GridCoords:
...
+        # These properties are identical for both u & v and defined here
+        self._df = 1 / (npix * self._dimage) * 1e-3  # [kλ]
```
iancze, line 72: I'm not a huge fan of `f` as a variable here. `u` and `v` are commonly used as spatial frequency variables in radio astronomy. `f` implies more about the observing wavelength and would be confusing to users. 

`k` could be a potential variable, but again, it's not really used in this way in radio astronomy. It would be great if you could find a variable that retains its interpretability as either a u or v coordinate.

src/mpol/coordinates.py
```diff
@@ -50,109 +51,208 @@ class GridCoords:
...
+        self._f_centers = self._df * (np.arange(npix) - npix // 2)
+        self._min_f = float(self._f_edges.min())
+        self._max_f = float(self._f_edges.max())
 
-        int_l_centers = np.arange(self.npix) - self.npix // 2
-        int_m_centers = np.arange(self.npix) - self.npix // 2
-        self.l_centers = self.dl * int_l_centers  # [radians]
-        self.m_centers = self.dm * int_m_centers  # [radians]
+        # max u or v freq supported by current grid
+        self.max_grid = get_max_spatial_freq(cell_size, npix)
```
iancze, line 79: A more descriptive name for this, perhaps like `max_uv_grid_value`

src/mpol/coordinates.py
```diff
@@ -50,109 +51,208 @@ class GridCoords:
...
+        ]  # [kλ]
+
+    # --------------------------------------------------------------------------
+    # Non-identical u & v properties
+    # --------------------------------------------------------------------------
+    @cached_property
+    def sky_u_centers_2D(self) -> NDArray[floating[Any]]:
+        # only useful for plotting a sky_vis
+        # uu increasing, no fftshift
+        return np.tile(self.u_centers, (self.ncell_u, 1))
```
iancze, line 181: What is the reason for choosing `tile` over `meshgrid`?

iancze (approved):
I think this looks good and is ready to merge, but would appreciate finding a different variable for spatial frequency than `f`, preferably one that retains interpretability as `u` or `v`.

src/mpol/coordinates.py
kadri-nizam: How about `_uv_pixel_width`?

src/mpol/coordinates.py
kadri-nizam: Sounds good!

src/mpol/coordinates.py
kadri-nizam: `meshgrid(x, y)` returns tiled `x` and `y.T` at the same time. With the properties, I'm defining only one of the attributes so used `tile` instead. If you think `meshgrid` feels more natural (and readable), I can pull the definitions of these attributes up to the constructor and define it that way.

iancze, line 181:
```suggestion
        # tile replicates the 1D u_centers array to a 2D array the size of the full UV grid
        return np.tile(self.u_centers, (self.ncell_u, 1))
```

iancze: OK I added a code comment so that it's more clear what tile is doing. I'm used to seeing meshgrid many places in the numpy codebase, so that was obvious to me. But I agree `tile` makes more sense here.

src/mpol/coordinates.py
iancze: Great! But I'm not sure it needs to be a private variable? I could see the user potentially wanting to access this in a few cases.

kadri-nizam: The user will access the hidden variable through the attributes `du` and `dv` which are in`@property` decorators.

iancze: Oh, yea. Looks good!

iancze:
Merged into WIP-v0.3 branch by updated #250 . Thanks for the contribution, @kadri-nizam !

PR #195, opened 2023-11-01
Single Zenodo record reference

### LINKED ISSUE #194: Condense Zenodo records references to single instance
For downloading large files used for docs and tests, there are multiple hardcoded references to the Zenodo record from which the files should be downloaded (2 in the codebase itself and several in the docs). It would be better to have the record as a single variable that can be imported.

Add a `zenodo_record` var somewhere it can be referenced within the test files, the jupyter notebooks in the docs, and the github actions workflows. 

**Describe alternatives you've considered**
Not keeping the Zenodo record updated in the docs, which works for now but could become an issue if the docs change.

**Additional context**

**How might you be able to help with the development of this feature?**
I'll create a PR to put `zenodo_record` in  `mpol.__init__` and reference it where needed.

jeffjennings:
Reduces the number of hardcoded Zenodo record refernces in the codebase and docs to one, in `__init__`. 

Maybe changing `tests.yml` as I have is not ideal for the cache(?), but it's what I could think of to be able to reference anything in the codebase at that point in the workflow. 

Closes #194

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -13,12 +13,11 @@ jobs:
         with:
           python-version: 3.8
       - name: Install dependencies needed to download files
-        # we're just installing the PyPI version of mpol here to reference the
-        # zenodo record number in __init__. below we'll reinstall the current 
-        # version of mpol for the tests below.
+        # we're just installing mpol here to reference the zenodo record number 
+        # in __init__. below we'll reinstall for the tests.
         run: |
           pip install astropy
-          pip install mpol
+          pip install .
       - name: Cache/Restore the .mpol folder cache
         uses: actions/cache@v3
         env:
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -13,8 +13,12 @@ jobs:
         with:
           python-version: 3.8
       - name: Install dependencies needed to download files
+        # we're just installing the PyPI version of mpol here to reference the
+        # zenodo record number in __init__. below we'll reinstall the current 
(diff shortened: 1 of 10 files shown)


.github/workflows/tests.yml
```diff
@@ -13,8 +13,11 @@ jobs:
         with:
           python-version: 3.8
       - name: Install dependencies needed to download files
+        # we're just installing mpol here to reference the zenodo record number
```
iancze, line 16: Can you explain why the zenodo reference number is needed at this point in the tests? Is it for the later `download_external_files.py` stage? I think this is a fine solution, but if you really wanted to defer installation of the package until the next stage of the workflow, you could modify the download files to take a path to a zenodo reference, and you would have a script that reads that path from the __init__.py source file (something similar to the [single source Python version](https://packaging.python.org/en/latest/guides/single-sourcing-package-version/) but this is probably overkill).

iancze (approved):
Looks good to me. If you're OK with the `tests.yml` layout then I'd say merge. Thanks!

.github/workflows/tests.yml
jeffjennings: Yes that's right, the reference number is needed for `download_external_files.py` to reference. It's a little clunky, but yeah I think it's fine. Your reference is a good one, but if you're ok with a bit of redundancy at this point in the workflow, I'm happy to keep the current workaround.

jeffjennings:
> Looks good to me. If you're OK with the `tests.yml` layout then I'd say merge. Thanks!

Thanks for the quick review!

PR #197, opened 2023-11-03
Add package test on push to main, update coverage badge

### LINKED ISSUE #196: Package test not running on main
**Describe the bug**
Not strictly a bug, but `tests.yml`, the package test, doesn't run on pushes to `main`. In case commits were made to `main` not through a PR, the test wouldn't run. 

Also,  the `package test` badge in the README and the docs is generated from `tests.yml`, so it shows the status of the last instance of `tests.yml` that was run, and thus can show a `failing` status (as it does now) for branches that aren't `main`. It should really just show the status of `main`, not experimental branches.

I'll create a PR to address these changes.

jeffjennings:
- Adds a trigger of `tests.yml` (the package test) for pushes to `main`. 
- Updates the `package test` badge in the readme and docs to show the status of the package test only on `main` (rather than the last time this test was run across all branches). 

Closes #196

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -1,6 +1,8 @@
 name: package test
 
 on:
+  push:
+    branches: [main]
   pull_request:
 
 jobs:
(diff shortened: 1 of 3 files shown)


iancze (approved):
Looks good, nice catch with the badge, that was bugging me...

PR #204, opened 2023-11-17
Hyperparameter tuning

jeffjennings:
Implements automated determination of regularizer strengths and learning rate within a cross validation loop using `ray tune`.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -58,6 +58,8 @@ EXTRA_REQUIRES = {
     ],
     "analysis": [
         "frank>=1.2.1",
+        "ray>=2.8.0",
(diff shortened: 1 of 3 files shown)


jeffjennings:
After messing with raytune for a while, I'm not convinced it's the best solution for regularizer strength and learning rate tuning. It requries a lot of overhead to convert training and crossval objects into ray format and adds big dependencies. Built-in pytorch schedulers can probably do much of the same without refactoring code. Raytune stack tracebacks are also pretty unhelpful, and their API seems to be changing frequently. I'm going to look into a simpler approach in pytorch and only come back to this if we need it. Will leave branch open for now.

Issue #63, opened 2021-06-01
HD 143006 Imaging Tutorial Part 3


iancze:
An overview of the HD 143006 imaging series is provided in #25

Part III is meant to be the "production-ready" version of a MPoL imaging script. The idea is to use computational resources to do hyperparameter sweeps. One example is using [Ray Tune](https://docs.ray.io/en/master/tune/index.html), though we might also want to suggest alternatives.

* Explore "production-ready" scripting layouts (no Jupyter notebooks). Some useful code and directory structures [here](https://github.com/pytorch/examples/tree/master/fast_neural_style). 
* We have a starting point in the [examples/HD143006](https://github.com/MPoL-dev/MPoL/tree/main/examples/HD143006) subdirectory. These need to be improved to be more generalizable (no hardcoded paths) similar to the examples in the Pytorch directory. We also need to document how to visualize loss, etc. with tensorboard.
~~* Explore thorough hyperparameter testing with Ray Tune.~~
* Develop new priors if necessary to achieve a good image

This "tutorial" document might not actually be a *.py file -> Jupytext -> ipynb like the other tutorials. Rather, it might just be a regular *.rst file with text describing how one might go about running the each of the steps. If necessary, we can include a final notebook that uses the optimal parameters to visualize the best-fit image.

jeffjennings:
Suggest adding a step demonstrating 'pre-training' to a dirty image (#206) to this tutorial.

jeffjennings:
Suggest adding mention of use of a learning rate scheduler (#207) to this tutorial.

jeffjennings:
Suggest adding the figure generated by #208 to this tutorial.

PR #206, opened 2023-11-29
Add utility for training to dirty image and integrate into cross-val

jeffjennings:
- Adds a function to train against a dirty image. 
- Optionally integrates this function into a cross-validation loop.
     * Training against a dirty image is run once at the start of cross-val, with the resulting state dictionary saved. The state dict is then loaded to update each new model at the start of the training loop for each k-fold.
- Adds a test of the function.

### DIFF
--- a/src/mpol/crossval.py
+++ b/src/mpol/crossval.py
@@ -12,7 +12,7 @@ from numpy.typing import NDArray
 
 from mpol.datasets import Dartboard, GriddedDataset
 from mpol.precomposed import SimpleNet
-from mpol.training import TrainTest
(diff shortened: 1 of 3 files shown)


iancze (approved):
Looks good, thanks. It would be helpful to have documentation (a tutorial, or other narrative text) explaining how one might start using these tools in an example workflow. Perhaps for HD 143006 part 3, or the IM Lup example.

jeffjennings:
Great, and good point - the HD 143006 part 3 tutorial sounds like a nice place. I've added a note in #63.

PR #208, opened 2023-11-29
Train diagnostic figure update

jeffjennings:
NOTE: This PR should only be reviewed after #207 and #206 are merged into `main` and `main` merged into this branch (it was branched from 207, which was branched from 206)

- Adds some new plots to the diagnostic figure that's generated at some chosen interval during a training loop:
    * difference image between model image at current iteration and that at the previous instance of the interval
    * flux of model image vs iteration
    * learning rate vs iteration
- Tracks model flux and learning rate in training loop
- Some general cleanup in the training and cross-val classes

### DIFF
--- a/src/mpol/crossval.py
+++ b/src/mpol/crossval.py
@@ -170,7 +170,7 @@ class CrossValidate:
         for kk, (train_set, test_set) in enumerate(split_iterator):
             if self._verbose:
                 logging.info(
-                    "\nCross-validation: k-fold {} of {}".format(kk, self._kfolds)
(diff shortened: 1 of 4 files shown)


jeffjennings:
Example of the new `train_diagnostics_fig`
entropy1e-05_TSV0 0001__train_diag_kfold0_epoch02623

jeffjennings:
Thanks! I've made a note about adding this in #63; there might be a good earlier place in the docs to put it, but #63 is looking like the hub for the next tutorial additions to the docs.

PR #209, opened 2023-11-29
Image residuals utility

jeffjennings:
Adds a function to quickly obtain an image of residual model visibilities and a cmap norm of the image for plotting.

Will be covered in tests by #211, which will call this function while producing a cross-val diagnostic figure.


jeffjennings:
Thanks for the review!

PR #210, opened 2023-11-29
Fits image processor

jeffjennings:
- Adds a new script for IO modules, and adds the `ProcessFitsImage` class to it. This class has fucntionality to intake a .fits image, load and return it as a numpy array, and read the image extent and beam to return those. 

- Adds a new test script with a test of the routines in `ProcessFitsImage`

I think this is the best approach to adding these functionalities into the codebase, as `beam_utils` isn't public, and anyway making it a dependency would inherit its `casatools` and similiar dependencies. I've very slightly modified the code for the functions from `beam_utils`. Most of this functionality is already in a tutorial, but the wrapping class in this PR is very convenient to load in a clean image for comparison. The class will be used in #211  to load a clean .fits image for comparison to cross-val MPoL results.

### DIFF
--- a/src/mpol/input_output.py
+++ b/src/mpol/input_output.py
@@ -2,3 +2,19 @@ import numpy as np
 
 from astropy.io import fits
 
+class ProcessFitsImage:
+    """
+    Utilities for loading and retrieving metrics of a .fits image
+
+    Parameters
+    ----------
+    filename : str
+        Path to the .fits file
+    channel : int, default=0
+        Channel of the image to access
+    """
+
+    def __init__(self, filename, channel=0):
+        self._fits_file = filename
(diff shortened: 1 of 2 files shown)


iancze (approved):
Sure, looks good. I think beam_utils is basically just a holding place for some potentially related code, with the ultimate goal that we'll want to delete it. Definitely do not want to inject any dependency on casatools or casatasks into the base MPoL package. This would immediately compromise the ability to install on Python 3.9, 3.10, 3.11 given current support and known bugs.

jeffjennings:
Ok cool I see, thanks for explaining beam_utils - I didn't want to steal from it. And yeah, big reasons not to add those dependencies. Thanks for the review!

PR #212, opened 2023-12-01
Dartboard: give train splits have short baseline data

jeffjennings:
- Alters `crossval.DartboardSplitGridded` to give every train split the cells belonging to the smallest baseline bin in the dartboard. Withholds these cells from all test sets. 
     - This resolves an issue in which the splitting can result in one k-fold performing quite different from the others (having an outlier loss function and CV score) because the split for that k-fold happened to put some of the smallest baseline cells in the test set. Thus it seems even missing some of the shortest baseline points can significantly affect the modeling.

- Sets the default `Dartboard` in `CrossValidate` to set `q` cells based on the max baseline in the data, not the max of the Fourier grid (which can be quite a bit larger). This ensures the points that every train cell gets aren't too large a fraction of the total vis distribution. I've set the `q` bins to be the same as in `vis_histogram_fig` so that the figure made by that corresponds to the splits in a cross-val loop by default (also added args to pass in `q` and `phi` bins to `CrossValidate`).

- Updates the figure produced by `plot.split_diagnostics_fig` to look a bit better

### DIFF
--- a/src/mpol/crossval.py
+++ b/src/mpol/crossval.py
@@ -295,6 +295,8 @@ class RandomCellSplitGridded:
     Treats `dataset` as a single-channel object with all data in `channel`.
 
     The splitting doesn't select (preserve) Hermitian pairs of visibilities.
+
+    All train splits have the highest 1% of cells by gridded weight
     """
 
     def __init__(self, dataset, k=5, seed=None, channel=0):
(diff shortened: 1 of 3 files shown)


iancze (approved):
This looks excellent, and will be a major improvement to the reliability of the cross val process. Thanks!

PR #213, opened 2023-12-01
Add a figure for 1d visibility distributions

jeffjennings:
~~NOTE: This should only be reviewed after #212 is merged into `main` and `main` into here.~~ [done]

- Adds a plotting function to generate a figure with 1d visibilities: observed, mpol model, residual. Plots both real and imaginary, can plot either projected or deprojected (if a geometry is supplied). Will be used to generate quick looks at model performance at end of cross-val loop.

### DIFF
--- a/src/mpol/plot.py
+++ b/src/mpol/plot.py
@@ -656,6 +656,64 @@ def image_comparison_fig(model, u, v, V, weights, robust=0.5,
 
 def vis_1d_fig(model, u, v, V, weights, geom=None, rescale_flux=False, 
               bin_width=20e3, title="", channel=0, save_prefix=None):
+    """
+    Figure for comparison of 1D projected MPoL model visibilities and observed 
+    visibilities. Plots:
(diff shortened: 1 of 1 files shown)


iancze (approved):
This looks really useful for quick looks and assessing convergence. Is the idea that there should be a one-to-one correspondence between binned observations and MPoL points? E.g., should the points at short baselines be matching, and is the MPoL point falling short indicative of a training problem?

jeffjennings:
Yes that's right, if they'res using the same binning there would be perfect correspondence in principle. E.g. since the points at short baseline aren't matching, it quickly indicates the model is underestimating the total flux. Or if there were a source that were highly asymmetric, seeing the structure in Im(V) and how well the model is tracing that could be interesting. I also looked at plotting the 2D residual visibilities, but haven't thought of a way to make it clear what's happening by eye (unless the fit is very bad), short of the imaged 2D residuals in the `image_comparison_fig` output.

PR #214, opened 2023-12-02
Add cross-val diagnostic figure

jeffjennings:
NOTE: This should only be reviewed after https://github.com/MPoL-dev/MPoL/pull/213 is merged into main and main into here.

- Adds the plotting function `plot.crossval_diagnostics_fig` to produce a diagnostic figure using a `CrossValidate` object. Figure shows a loss evolution and CV score for each k-fold.

- Sets `train.TrainTest.trainloss_lambda_guess` to only run if 1+ of the regularizers used has `guess` of `True` (user-provided).

### DIFF
--- a/src/mpol/crossval.py
+++ b/src/mpol/crossval.py
@@ -280,6 +280,11 @@ class CrossValidate:
         return self._cv_score
 
     @property
+    def split_method(self):
+        """String of the method used to split the dataset into train/test sets"""
+        return self._split_method
+    
+    @property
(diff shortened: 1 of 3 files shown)


jeffjennings:
Example figure: 
IMLup_crossval_diagnostics

iancze (approved):
Looks very helpful -- the pipeline is really coming together! It's interesting how similar the loss profiles are for the different K-folds (though I guess this is log scale).

PR #216, opened 2023-12-02
Revert "Add cross-val diagnostic figure"

jeffjennings:
Reverts MPoL-dev/MPoL#214


jeffjennings:
I merged #214 before #213, and #214 was branched off #213. I guess it doesn't really matter, but I'd rather do it correctly.

PR #220, opened 2023-12-04
1D analysis figure

jeffjennings:
- Adds figure for radial brightness profile analysis, given a user-supplied geometry. Meant primarily as a pipeline output.
- Cleans up some misc small things related to this figure or the pipeline

Issue #219, opened 2023-12-03
mpol.plot.vis_histogram_fig to accept loose visibilities?


iancze:
**Is your feature request related to a problem or opportunity? Please describe.**
The new `mpol.plot.vis_histogram_fig` routine is great! But it requires a `mpol.datasets.GriddedDataset` object as input, which requires the user to make assumptions about the image size via a `GridCoords` object so that they can use the `DataAverager` to produce a `GriddedDataset`. They're probably going to want to do this at somepoint anyway, but it seems a bit premature to make them do this while they're still familiarising themselves with the visibility dataset. 

The current docs also have some `unit=:math:[klambda]` that didn't render.

**Describe the solution you'd like**
Could `vis_histogram_fig` accept (non-Hermitian pairs) `uu` and `vv` numpy arrays directly, as `DataAverager` or `DirtyImager` do? Could also borrow logic from inside `DirtyImager` to expand the arrays so that they properly include Hermitian pairs in the histogram.

And then the logic inside `vis_histogram_fig` could be simplified slightly since we don't need to worry about indexing zero or non-zero `GriddedDataset` cells.

jeffjennings:
It's nice to make utility functions more flexible, though in this case I think it's a bit overkill. It is just a plotting function, the coarse and fine 1d histograms produced by it with the gridded data represent the visibility distribution well enough for practical purposes, and if a user doesn't know their preferred image size (which they really should), a simple conversion of `min(np.hypot(u, v))`, or `max` for pixel size, would suffice. `utils.get_optimal_image_properties` also gives a sensible pixel size given a desired image width and the (u,v) points. 

I think it's useful and convenient to keep support for a gridded dataset as input, and the simplified logic of switching wouuld amount to removing one line (with more added to handle loose vis). I think a simpler solution is a small plotting function that shows the 2d (u,v) distribution, #220

PR #221, opened 2023-12-04
uv distribution figure

jeffjennings:
~~NOTE: Should be reviewed/merged after #220 (was branched from).~~ [done] 

Adds a plotting function to show the 2d (u,v) distribution, with points colored by Re(V).

Meant to address #219

### DIFF
--- a/src/mpol/plot.py
+++ b/src/mpol/plot.py
@@ -165,6 +165,30 @@ def plot_image(image, extent, cmap="inferno", norm=None, ax=None,
 
 
 def uv_dist_fig(u, v, V, cmap="inferno", title="", save_prefix=None):
+    """
+    Figure showing the (u,v) distribution of a dataset, with points colored by
+    the Re(V) amplitude.
+
+    Parameters
+    ----------
+    u, v : array, optional, unit=[k\lambda], default=None
+        Data u- and v-coordinates
+    V : array, optional, unit=[Jy], default=None
+        Data visibility amplitudes
+    cmap : str, default="inferno"
+        Colormap
+    title : str, default=""
+        Figure super-title
+    save_prefix : string, default = None
+        Prefix for saved figure name. If None, the figure won't be saved
+
+    Returns
+    -------
+    fig : Matplotlib `.Figure` instance
+        The generated figure
+    ax : Matplotlib `~.axes.Axes` class
+        Axis of the generated figure
+    """
 
     fig, ax = plt.subplots()    
     fig.suptitle(title)
@@ -184,6 +208,8 @@ def uv_dist_fig(u, v, V, cmap="inferno", title="", save_prefix=None):
     plt.close()
 
     return fig, ax
+    
+
 def vis_histogram_fig(dataset, bin_quantity='count', bin_label=None, q_edges=None, 
     phi_edges=None, q_edges1d=None, show_datapoints=False, save_prefix=None):
     r"""
--- a/src/mpol/plot.py
+++ b/src/mpol/plot.py
@@ -164,6 +164,26 @@ def plot_image(image, extent, cmap="inferno", norm=None, ax=None,
     return im, cbar
 
 
+def uv_dist_fig(u, v, V, cmap="inferno", title="", save_prefix=None):
+
+    fig, ax = plt.subplots()    
+    fig.suptitle(title)
+
+    scat = plt.scatter(u / 1e3, v / 1e3, s=0.1, marker='.', c=V.real * 1e3, 
+                       cmap=cmap, rasterized=True)
+    cbar = plt.colorbar(scat, ax=ax, location="right", pad=0.1)
+
+    cbar.set_label("Re(V) [mJy]")
+
+    ax.set_xlabel(r'u [M$\lambda$]')
+    ax.set_ylabel(r'v [M$\lambda$]')
+    
+    if save_prefix is not None:
+        fig.savefig(save_prefix + '_uv_dist.png', dpi=300)
+    
+    plt.close()
+
+    return fig, ax
 def vis_histogram_fig(dataset, bin_quantity='count', bin_label=None, q_edges=None, 
     phi_edges=None, q_edges1d=None, show_datapoints=False, save_prefix=None):
     r"""


jeffjennings:
Example figure generated: 
IMLup_uv_dist

iancze:
Thanks. I'm curious about your choice to plot `Re` only? I think amplitude would be more generally useful.

I think the functionality that I was searching for in #219 is a way to roughly estimate the signal (binned amplitude) and sensitivity (binned weights) from the visibilities as demonstrated in the figure in #97, which can be generated without any imaging parameters by doing the histogram binning directly on the loose visibilities. In terms of an analysis workflow, I would expect one to first plot up the baseline distribution (as in this #221, perhaps, even just using u,v as scatter points with no amp information), then maybe plot a 1D distribution of amplitudes, and examine 2D sensitivity & num visibilities using a histogram (since it's hard to do from a scatter plot like #221, so many points overlap), and then move on to gridding and making a dirty image, etc.

Is the recommended way to get here then to choose imaging parameters, create a GriddedDataset, and then feed that into `mpol.plot.vis_histogram_fig`? It seems like this route might hamper one from estimating number of visibilities and other metrics like scatter relative to expected scatter. I guess there's also a consideration of plotting histograms on a cartesian grid (more relevant for scatter relative to weights) vs. a polar grid (easier to visualize sensitivity).

jeffjennings:
Sure, the coloring could be amplitude or Re(V) or Im(V) or none. 

I understand what you had in mind, though the reason I wanted to leave `vis_histogram_fig` as-is is that it's showing what the model sees, which is useful especially when using the dartboard to split train/test. It's a part of the modeling pipeline in that sense.

I'm less clear on where this kind of initial data visualization / pre-processing should go. It's in the [docs](https://mpol-dev.github.io/MPoL/ci-tutorials/gridder.html) a bit, but also in the [visread docs](https://mpol-dev.github.io/visread/tutorials/rescale_AS209_weights.html#checking-scatter-for-each-spectral-window). I'm not sure if you have in mind to use a minimal-dependency visread to pre-process and visualize data that can then be fed to MPoL, or if you want MPoL to do it all. If the latter, I think the routines in this kind of 'pre-processing pipeline' should be distinct from those used in the modeling pipeline to avoid confusion and making things kind of complicated.

In any case, this PR doesn't have to be merged. It was meant to be a standalone pre-processing routine, but maybe it doesn't belong in MPoL. I do think #219 should be rescoped though, as to me at least it's better if the routines going into these two pipelines are clearly separated (either in MPoL or between MPoL and visread).

iancze:
Thanks, you make very good points that suggest some organizational thought about the purposes of the visread v. MPoL packages. 

My original thinking for the bifurcation of the packages was to have `visread` with the `casatools` dependency for reading/averaging/exporting visibilities from a measurement set, and MPoL for everything related to RML imaging (especially those depending on PyTorch). But recent developments complicate this picture a little bit.

I'm currently in the process of reorganizing `visread` to have two install tiers: one with `casatools` dependency to do the reading from an MS, and another without `casatools` dependencies, so routines to estimate visibility scatter and do channel broadcasting can be easily imported into modern environments that might read from an `.asdf` file, since these seem misplaced in an imaging-focused MPoL package.

* This separation makes `visread` (w/o CASA deps) a natural fit for routines like the uv distribution figure, 1D histograms, etc, since these don't involve PyTorch or anything specific to RML.
* On the flipside, though, if we can isolate the `casatools` to an optional dependency, why not just incorporate *all* of the visread functionality into MPoL itself, ms reading included? Would this be easier for the user to deal with? We would essentially have a suite of 'data exploration' tools within MPoL that can average / plot visibilities, and we would have a (`casatools`-dependent) module that could read visibilities from an MS. 

Thinking through how a user might export visibilities from a measurement set for MPoL analysis (e.g., using my current tools available):

1) in a Python 3.8 environment installed on RHEL (i.e., the best I have available where I can actually install `casatools` without error). *currently*: install `visread`, read and average visibilities, write to a neutral data format like .npy or .asdf. *anticipated*: install `mpol[casa]`, read and average visibilities, write to neutral data format. Then:
2) in a modern dev environment (for me, > Python 3.10, CUDA docker image): run mpol after loading .npy or .asdf file. 

Is it confusing to recommend to someone that they install MPoL two different ways on two different Python versions? Is this preferable to the current setup, where in the MPoL docs we may frequently recommend (if not downright require) the user to also install visread to do some visibility manipulations?

Of course, if a user's modern dev environment *is* Python 3.8 RHEL, then they would just use `mpol[casa]` directly. The perennial problem is that the `casatools` package significantly lags modern python releases, so making this a core requirement (`mpol`) substantially constraints the install environments and blocks a large user base from using this software, which we do not want to do.

jeffjennings:
> Thanks, you make very good points that suggest some organizational thought about the purposes of the visread v. MPoL packages.
> 
> My original thinking for the bifurcation of the packages was to have `visread` with the `casatools` dependency for reading/averaging/exporting visibilities from a measurement set, and MPoL for everything related to RML imaging (especially those depending on PyTorch). But recent developments complicate this picture a little bit.
> 
> I'm currently in the process of reorganizing `visread` to have two install tiers: one with `casatools` dependency to do the reading from an MS, and another without `casatools` dependencies, so routines to estimate visibility scatter and do channel broadcasting can be easily imported into modern environments that might read from an `.asdf` file, since these seem misplaced in an imaging-focused MPoL package.
> 
> * This separation makes `visread` (w/o CASA deps) a natural fit for routines like the uv distribution figure, 1D histograms, etc, since these don't involve PyTorch or anything specific to RML.
> * On the flipside, though, if we can isolate the `casatools` to an optional dependency, why not just incorporate _all_ of the visread functionality into MPoL itself, ms reading included? Would this be easier for the user to deal with? We would essentially have a suite of 'data exploration' tools within MPoL that can average / plot visibilities, and we would have a (`casatools`-dependent) module that could read visibilities from an MS.

You could have both -- all of this functionality is in visread, and there's also a `pre-processing` pipeline in MPoL that just calls several visread routines to e.g. make plots or output metrics.

> 
> Thinking through how a user might export visibilities from a measurement set for MPoL analysis (e.g., using my current tools available):
> 
> 1. in a Python 3.8 environment installed on RHEL (i.e., the best I have available where I can actually install `casatools` without error). _currently_: install `visread`, read and average visibilities, write to a neutral data format like .npy or .asdf. _anticipated_: install `mpol[casa]`, read and average visibilities, write to neutral data format. Then:
> 2. in a modern dev environment (for me, > Python 3.10, CUDA docker image): run mpol after loading .npy or .asdf file.
> 
> Is it confusing to recommend to someone that they install MPoL two different ways on two different Python versions? Is this preferable to the current setup, where in the MPoL docs we may frequently recommend (if not downright require) the user to also install visread to do some visibility manipulations?

Yes I think it's confusing and messy to have to install MPoL two ways in two Python environments. Requiring visread[non-casa] as a dependency seems perfectly fine - it's light and its dependencies are already required by MPoL (["numpy", "scipy", "astropy"]).

> 
> Of course, if a user's modern dev environment _is_ Python 3.8 RHEL, then they would just use `mpol[casa]` directly. The perennial problem is that the `casatools` package significantly lags modern python releases, so making this a core requirement (`mpol`) substantially constraints the install environments and blocks a large user base from using this software, which we do not want to do.

Definitely not. That would add a lot of headaches for devs and users.

> 
> After writing out the use case, I think it's preferable to push 'data exploration' capabilities into visread (minus casatools dep) and keep MPoL focused on the RML imaging + data modeling (e.g., Pyro + SVI). It's easier to explain to the user that they can use visread to prepare and examine the data (and that this may require a `casatools` dependency for reading from an MS) than to explain that they need to install MPoL twice into two different environments, but curious to hear your thoughts @jeffjennings .

I agree. Again the docs can point to using visread for data prep/examination, and these functionalities can be run in a data processing pipeline (wrapper) in MPoL.

jeffjennings:
Closing as out of scope with v0.3 redesign

PR #222, opened 2023-12-04
Dartboard: handle case for 1 k-fold

jeffjennings:
`DartboardSplitGridded` doesn't currently handle the case for 1 k-fold, only 2+. `DartboardSplitGridded` isn't needed to train on a full dataset obviously, but for a standard train/test procedure, PR adds default behavior in `DartboardSplitGridded` of splitting dataset into (roughly) 80/20 train/test when k-fold == 1. The train set still keeps the smallest dartboard cells, as is does for 2+ k-folds.


iancze (commented):
Looks great, thanks. Are there any relevant tests that could also be included here? For example, give it a mock dataset and see if the splits are in the ratio expected?

jeffjennings:
Test added.

PR #230, opened 2023-12-08
Change DirtyImager to store minimal visibility data, Hermitian conjugates computed on the fly via @property

### LINKED ISSUE #224: Memory issues in DataAverager and DirtyImager
**Describe the bug**
Within `DataAverager` and `DirtyImager`, when `check_visibility_scatter` is True in `DataAverager.to_pytorch_dataset` or `DirtyImager.get_dirty_image`, `GridderBase._check_scatter_error` is called, thus `GridderBase.estimate_cell_standard_deviation`.  With large datasets, specifically the test dataset for IM Lup in .asdf, `estimate_cell_standard_deviation` causes memory crash on a system with 8 GB of VRAM (I'm running things on the GPU). If running on a CPU (with 16 GB free RAM), the code doesn't crash (but is much slower).

**Suggested fix**
A quick look at this function suggests its operations could be batched,  working on smaller chunks of the loose visibilities in sequence. 

**Desktop (please complete the following information):**
 - OS: WSL - Ubuntu
 - Python Version: 3.9.16
 - MPoL Version: 0.2.0 (+ current commits to main)

iancze:
Closes #224 . 

For the IM Lup DSHARP dataset (non-channel averaged) size 1.1 Gb, this results in a DirtyImager with only a 3.8 Gb memory footprint, compared to 5.1 Gb previously. 

This also reduced peak memory usage in Jeff's script in #224 from ~8 Gb to 6.9 Gb.

### DIFF
--- a/src/mpol/gridding.py
+++ b/src/mpol/gridding.py
@@ -597,21 +597,39 @@ class DirtyImager(GridderBase):
         # make sure we still fit into the grid
         self.coords.check_data_fit(uu, vv)
 
-        # expand the vectors to include complex conjugates
-        uu_full = np.concatenate([uu, -uu], axis=1)
-        vv_full = np.concatenate([vv, -vv], axis=1)
+        # initialize base objects 
+        # Hermitian conjugates taken care of by @property below
+        self.uu_base = uu
+        self.vv_base = vv
+        self.weight_base = weight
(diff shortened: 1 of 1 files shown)


jeffjennings (approved):
Looks good

PR #235, opened 2023-12-10
Docstrings: docs API rendering

### LINKED ISSUE #234: Docstrings do not fully render for several mpol.plot routines
**Describe the bug**
The docstrings for some routines in mpol.plot do not render correctly.

When building the docs, I get 
```
/Users/ian/Documents/Research/Disks/RML/MPoL/src/mpol/plot.py:docstring of mpol.plot.vis_1d_fig:20: ERROR: Unexpected indentation.
/Users/ian/Documents/Research/Disks/RML/MPoL/src/mpol/plot.py:docstring of mpol.plot.vis_1d_fig:35: ERROR: Unexpected indentation.
/Users/ian/Documents/Research/Disks/RML/MPoL/src/mpol/plot.py:docstring of mpol.plot.get_residual_image:10: WARNING: Block quote ends without a blank line; unexpected unindent.
/Users/ian/Documents/Research/Disks/RML/MPoL/src/mpol/plot.py:docstring of mpol.plot.radial_fig:5: ERROR: Unexpected indentation.
/Users/ian/Documents/Research/Disks/RML/MPoL/src/mpol/plot.py:docstring of mpol.plot.get_image_cmap_norm:1: WARNING: Block quote ends without a blank line; unexpected unindent.
/Users/ian/Documents/Research/Disks/RML/MPoL/src/mpol/plot.py:docstring of mpol.plot.radial_fig:13: ERROR: Unexpected indentation.
/Users/ian/Documents/Research/Disks/RML/MPoL/src/mpol/plot.py:docstring of mpol.plot.radial_fig:38: ERROR: Unexpected indentation.
/Users/ian/Documents/Research/Disks/RML/MPoL/src/mpol/plot.py:docstring of mpol.plot.get_residual_image:13: WARNING: Block quote ends without a blank line; unexpected unindent.
```

This is with a fresh `pip install -e '.[dev]'` off of `main` and after doing `make clean`.  The render failure can also be seen live, e.g. for [`vis_1d_fig`](https://mpol-dev.github.io/MPoL/api.html#mpol.plot.vis_1d_fig), the intended list of plots seen in the source code

```
    Figure for comparison of 1D projected MPoL model visibilities and observed 
    visibilities. Plots:
    - Re(V): observed and MPoL model (projected unless `geom` is supplied)
    - Residual Re(V): observed - MPoL model (projected unless `geom` is supplied)
    - Im(V): observed and MPoL model (projected unless `geom` is supplied)
    - Residual Im(V): observed - MPoL model (projected unless `geom` is supplied)
```
renders as 

```
Figure for comparison of 1D projected MPoL model visibilities and observed visibilities. Plots: - Re(V): observed and MPoL model (projected unless geom is supplied) - Residual Re(V): observed - MPoL model (projected unless geom is supplied) - Im(V): observed and MPoL model (projected unless geom is supplied) - Residual Im(V): observed - MPoL model (projected unless geom is supplied)
```

**To reproduce**
Run docs build with current main branch `ff7e4bd`

**Expected behavior**
The docs build does not report errors and the live docs render correctly.

**Screenshots**
Screenshot 2023-12-09 at 23 07 54

**Suggested fix**
Convert these docstrings to follow the [napoleon format](https://sphinxcontrib-napoleon.readthedocs.io/en/latest/) as specified in `docs/conf.py` and used by the other routines in MPoL.

jeffjennings:
- Closes #234. Does so with some indents (rather than changing from the numpy docstring format to napoleon) because I don't find napoleon readable for long docstrings when looking at the code. I checked the API with `make clean` --> `make html`.

- Adds `mpol.onedim` to the docs API.

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -68,6 +68,12 @@ Cross-validation
 .. automodule:: mpol.crossval
 
 
+Analysis
+--------
+
+.. automodule:: mpol.onedim
+
+
 Plotting
 --------
 
--- a/src/mpol/onedim.py
+++ b/src/mpol/onedim.py
@@ -101,7 +101,7 @@ def radialV(fcube, geom, rescale_flux, chan=0, bins=None):
         for the difference between the inclined (observed) brightness and the 
         assumed face-on brightness, assuming the emission is optically thick. 
         The source's integrated (2D) flux is assumed to be:
-            :math:`F = \cos(i) \int_r^{r=R}{I(r) 2 \pi r dr}`.
+        :math:`F = \cos(i) \int_r^{r=R}{I(r) 2 \pi r dr}`.
         No rescaling would be appropriate in the optically thin limit. 
     chan : int, default=0
         Channel of the image cube corresponding to the desired image        
--- a/src/mpol/plot.py
+++ b/src/mpol/plot.py
@@ -888,12 +888,13 @@ def radial_fig(model, geom, u=None, v=None, V=None, weights=None, dist=None,
                channel=0, save_prefix=None):
     """
     Figure for analysis of 1D (radial) brightness profile of MPoL model image,
-    using a user-supplied geometry. Plots:
-    - MPoL model image
-    - 1D (radial) brightness profile extracted from MPoL image 
-      (supply `dist` to show second x-axis in [AU])
-    - Deprojectd Re(V): binned MPoL model and observations (if u, v, V, weights supplied)
-    - Deprojected Im(V): binned MPoL model and observations (if u, v, V, weights supplied)
+    using a user-supplied geometry. 
+    
+    Plots:
+        - MPoL model image
+        - 1D (radial) brightness profile extracted from MPoL image (supply `dist` to show second x-axis in [AU])
+        - Deprojectd Re(V): binned MPoL model and observations (if u, v, V, weights supplied)
+        - Deprojected Im(V): binned MPoL model and observations (if u, v, V, weights supplied)
 
     Parameters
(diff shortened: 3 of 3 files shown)


iancze:
Thanks for these changes, however, I think it is important to use the same docstring style throughout the code. It's more mental effort for the developer to maintain and write correct documentation when we have to switch between formats depending on the function or class. And sphinx seems to have some cracks through which doc bugs fall through when docstrings are mixed format (side-note, is there a more strict version of sphinx build we can use to catch these? I thought we were already doing this).

So I would either suggest we stick with napoleon or modify everything to numpy docstring format (a considerable task). Can you elaborate on what's difficult to read with napoleon? I find we have some fairly long docstrings that still (to me, at least) look good in code and napoleon is certainly more concise to write, which are the reasons I originally chose it over numpy format.

jeffjennings:
Yes it's just that when using VSCode or similar, reading the docstring for something like `fourier.NuFFT` is annoying because the line length far exceeds the recommended Python line length and requires a lot of scrolling. Several of my docstrings have long lines, and it looks like there are ~50 of them with the numpy docstring format. Since the docs render correctly, I'd prefer to focus on what I think are much more urgent issues than re-format all of these.

iancze:
OK, definitely agree this is not a high priority use of your time :) But I think we should converge on a single docstring style for the codebase as a longer term goal. 

I use VSCode too and the docstrings render fine for me (virtual WordWrap set to on), which seems necessary for editing most kinds of documents anyway.

